PR #2, opened 2019-09-19
scipion-em as a python module

fonsecareyna82:
* Converting scipion-em as a python module
  -  Added the necessary files
* Refactor all code
  - Fixed all imports

### DIFF
--- a/pwem/convert/image_handler.py
+++ b/pwem/convert/image_handler.py
@@ -335,7 +335,7 @@ class ImageHandler(object):
             n = inputSet.getSize()
             if n:
                 imageIter = iter(inputSet)
-                img = imageIter.next()
+                img = next(imageIter)
                 avgImage = self.read(img)
                 
                 for img in imageIter:
--- a/pwem/tests/data/test_data.py
+++ b/pwem/tests/data/test_data.py
@@ -21,13 +21,6 @@ import pwem.convert as emconv
 import pwem.objects as emobj
 import pwem.metadata as md
 
-# try:
-#     from xmipp3.convert import *
-# except:
-#     Domain._pluginNotFound('xmipp', doRaise=True)
-
-#xmippConvert = Domain.importFromPlugin('xmipp3.convert', doRaise=True)
-
 # set to true if you want to check how fast is the access to
 # the database
 SPEEDTEST = True
--- a/pwem/convert/__init__.py
+++ b/pwem/convert/__init__.py
@@ -26,7 +26,7 @@
 
 from .utils import getSubsetByDefocus, downloadPdb
 from .image_handler import ImageHandler, DT_FLOAT
-from .atom_struct import AtomicStructHandler, toCIF
+from .atom_struct import AtomicStructHandler, toCIF, cifToPdb
 from .headers import Ccp4Header
 from .symmetry import getSymmetryMatrices, Icosahedron
 from .transformations import identity_matrix
--- a/pwem/convert/sequence.py
+++ b/pwem/convert/sequence.py
@@ -49,7 +49,7 @@ UNAMBIGOUS_RNA_ALPHABET = 4
 from Bio.Seq import Seq
 from Bio.Alphabet import IUPAC
 from Bio import Entrez, SeqIO
-import urllib, sys
+import sys
 from Bio.SeqRecord import SeqRecord
 from Bio.Align.Applications import ClustalOmegaCommandline, MuscleCommandline
 from Bio import pairwise2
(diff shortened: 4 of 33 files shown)


pwem/__init__.py
```diff
@@ -33,11 +33,9 @@
 from protocol import *
 from convert import *
 from wizards import *
-from pyworkflow.utils import importFromPlugin
-from pyworkflow.wizard import Wizard
 from pyworkflow.viewer import Viewer
-from pyworkflow import Config
-
+from pyworkflow.wizard import Wizard
+from filehandlers import *
```
delarosatrevin: Maybe filehandlers should be organized into an existing submodule?
In Py3 you need to make relative import relative: e.g:
```python
from .filehandlers import *
```
Finally, import * are not a good idea anyway.

pwem/convert/image_handler.py
```diff
@@ -38,7 +39,7 @@
   sys.exit(-1)
 
 import pyworkflow.utils as pwutils
-from pyworkflow.em.constants import *
+from pwem.constants import *
```
delarosatrevin: what about ```import pwem.constants as cts``` and use it? It will make more clear where constants come from.

pwem/convert/image_handler.py
```diff
@@ -78,7 +79,7 @@ def fixXmippVolumeFileName(cls, image):
         """
         # We can not import Volume from top level since
         # data depends on this module
-        from pyworkflow.em.data import Volume, Movie
+        from pwem.data import Volume, Movie
```
delarosatrevin: Import should be moved at the top. We could also keep the prefixes

pwem/convert/symmetry.py
```diff
@@ -27,11 +27,11 @@
 This module returns the matrices related with the different
 point symmetries. Code based on chiomera file sym.py
 """
-from pyworkflow.em.constants import SYM_CYCLIC, SYM_DIHEDRAL, SYM_OCTAHEDRAL, \
-    SYM_TETRAHEDRAL, SYM_TETRAHEDRAL_Z3, SYM_I222, SYM_I222r, \
-    SYM_In25, SYM_In25r
+from pwem.constants import (SYM_CYCLIC, SYM_DIHEDRAL, SYM_OCTAHEDRAL,
```
delarosatrevin: Maybe we should agree in a prefix and use. I find this kind of imports no so clean.
Moreover, it is good to know in the code where things come from.

pwem/data.py
```diff
@@ -898,8 +899,7 @@ def __init__(self, filename=None, pseudoatoms=False, **kwargs):
 class EMSet(Set, EMObject):
 
     def _loadClassesDict(self):
-        import pyworkflow.em as em
-        classDict = em.Domain.getObjects()
+        classDict = Domain.getObjects()
```
delarosatrevin: There is an even more general function Domain.getMapperDict, that should be used for setting the mapper dict. 
With this new approach, the basic objects defined in objects.py are already there. So one can remove the `from pyworkflow.object import *` at the top here or the globals()

pwem/protocol/protocol_batch.py
```diff
@@ -29,14 +29,14 @@
 
 from pyworkflow.protocol.params import PointerParam, \
```
delarosatrevin: Please avoid `\` to split lines, it is not PEP8 compliant.

pwem/tests/data/test_data.py
```diff
@@ -9,10 +9,16 @@
 import pyworkflow.utils as pwutils
 import sqlite3
 from pyworkflow.utils import Timer
-try:
-    from xmipp3.convert import *
-except:
-    pwutils.pluginNotFound('xmipp', doRaise=True)
+from pwem.data import *
```
delarosatrevin: Let's try to avoid `import *` as much as possible.

setup.py
```diff
@@ -0,0 +1,38 @@
...
+# Arguments marked as "Required" below must be included for upload to PyPI.
+# Fields marked as "Optional" may be commented out.
+
+setup(
+    name='scipion-em',  # Required
+    version='1.0.0',  # Required
+    description='This modules contains classes related with EM',  # Required
+    long_description=long_description,  # Optional
+    url='https://github.com/scipion-em/scipion-em',  # Optional
+    author='I2PC',  # Optional
```
delarosatrevin: Not only I2PC is working on this ;)
There are some other classifiers that maybe is good to include as well. 
https://github.com/scipion-em/scipion-pyworkflow/blob/go-py3/setup.py#L115

PR #6, opened 2019-10-23
Ccp4Header --> consider path :mrc "decoration"

pconesa:
ImageHandler getDimensions: use ccp4header when possible
Ghost changes to reach further without xmipp.

### DIFF
--- a/pwem/convert/image_handler.py
+++ b/pwem/convert/image_handler.py
@@ -261,11 +261,20 @@ class ImageHandler(object):
             location = self._convertToLocation(locationObj)
             fn = location[1]
             ext = pwutils.getExt(fn).lower()
-            
+
+            # Local import to avoid import loop between ImageHandler and Ccp4Header.
+            from pwem.convert import Ccp4Header
+
             if ext == '.png' or ext == '.jpg':
                 im = PIL.Image.open(fn)
                 x, y = im.size # (width,height) tuple
                 return x, y, 1, 1
+
+            elif Ccp4Header.getFileFormat(fn) != Ccp4Header.UNKNOWNFORMAT:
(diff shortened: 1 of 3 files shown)


pconesa:
the name of the branch is not relevant.

pwem/convert/image_handler.py
```diff
@@ -261,11 +261,20 @@ def getDimensions(self, locationObj):
             location = self._convertToLocation(locationObj)
             fn = location[1]
             ext = pwutils.getExt(fn).lower()
-            
+
+            # Local import to avoid import loop between ImageHandler and Ccp4Header.
+            from pwem.convert import Ccp4Header
```
delarosatrevin: I think this import should be moved to the top...and also converted to a local import (since it is internal in the submodule):
```python
from .headers import Ccp4Header
```

pwem/xmipp-ghost/xmippLib.py
```diff
@@ -1061,14 +1061,27 @@ def MetaData():
...
-    pass
+class FileName:
+    """ Try to implement some basic code to reach further without xmipp"""
+    imageExtentions = []
+
+    def __init__(self, path):
+        self.path = path
+
+    def isImage(self):
+        return True
```
delarosatrevin: Everything is an image?

pwem/xmipp-ghost/xmippLib.py
```diff
@@ -1061,14 +1061,27 @@ def MetaData():
 
 MetaDataInfo = None
 
-def Image():
-    pass
+HEADER = None
 
+class Image:
```
delarosatrevin: We should think in the more general "backend" approach...but I'm fine with the change in the meantime.

PR #10, opened 2019-11-15
Changes to run scipion on windows

pconesa:


azazellochg:
you kidding, why do we support this abomination of the OS?

pconesa:
Is not that we will make a huge effort, probably not anymore than this, but
is good to be able to test how portable is scipion.

PR #23, opened 2020-01-08
fix html report generation

pconesa:


pwem/__init__.py
```diff
@@ -83,5 +83,4 @@ def defineBinariesMaxit(cls, default, env):
         # requirements bison, flex, gcc
 
 
-Domain.registerPlugin(__name__)
```
azazellochg: why this is gone? all pwem protocols fail now!

pconesa: Need an entry point and be installed again

JorMaister:
@azazellochg 
I clicked by error --> JorMaister requested a review from azazellochg
My apologies

PR #28, opened 2020-01-12
remove gempicker, ctffind3, replace grigoriefflab tests

azazellochg:


pwem/tests/workflows/test_workflow_mixed.py
```diff
@@ -81,7 +81,7 @@ def test_workflow(self):
         ProtCTFFind = Domain.importFromPlugin('cistem.protocols',
                                               'CistemProtCTFFind', doRaise=True)
         protCTF = self.newProtocol(ProtCTFFind, numberOfThreads=4,
```
pconesa: Is this still valid, using ProtCTFFind instead of CistemProtCTFFind?

pconesa: Forget it! i See it actually is cistem's one.

PR #29, opened 2020-01-19
add new output for relion 31

azazellochg:
This is required for new output type from relion motioncor - power spectra


delarosatrevin:
Hum...I don't link much to add more stuff to the base class, that is already pretty cluttered. I would prefer to consider a way to achieve the same with some implementation in the derived class, rather than the base one. @azazellochg we can further discuss about it.

delarosatrevin:
We are dealing with this output in a different way in the Relion protocol.

PR #34, opened 2020-01-21
Simplify chimera launch

pconesa:


pconesa:
This, for now allow launching Chimera from ShowJ that was broken. I removed the complexity by not having any client-server connection until we rescue this behavior once fixed.

setup.py
```diff
@@ -7,7 +7,7 @@
 
 # Always prefer setuptools over distutils
 from pwem import EM_PROGRAM_ENTRY_POINT, CHIMERA_ENTRY_POINT, \
```
delarosatrevin: This import with \ is not pep8 compliant.
I think we can just import pwem and use pwem.XXX as prefix of the constants.

fonsecareyna82: I'll fix in devel-fixes

PR #40, opened 2020-01-25
remove specific attrs

### LINKED ISSUE #30: deprecate old properties for ctf resolution
I think it's time to get rid of software-specific labels:

- _ctffind4_ctfResolution
- _gctf_ctfResolution
- _xmipp_ctfCritMaxFreq
- _ctffind4_crossCorrelation
- _gctf_crossCorrelation
- _ctffind4_ctfPhaseShift
- _gctf_ctfPhaseShift

They have been replaced by _phaseShift, _resolution and _fitQuality

azazellochg:
fix #30

### DIFF
--- a/pwem/objects/data.py
+++ b/pwem/objects/data.py
@@ -133,37 +133,24 @@ class CTFModel(EMObject):
         self._fitQuality = Float()
 
     def __str__(self):
-        if self._resolution.hasValue():
-            phaseShift = self.getPhaseShift() if self.hasPhaseShift() else 0
-            ctfStr = "defocus(U,V,a,psh,re,fit) = " \
-                     "(%0.2f,%0.2f,%0.2f,%0.2f,%0.2f,%0.2f)" % \
-                     (self._defocusU.get(),
-                      self._defocusV.get(),
-                      self._defocusAngle.get(),
-                      phaseShift,
-                      self._resolution.get(),
-                      self._fitQuality.get()
-                      )
-        else:   # TODO; remove eventually,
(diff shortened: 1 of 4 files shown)


pconesa:
Thank's @azazellochg , I'm actually not sure if we can remove them.

Old projects will not load....Old I guess are projects generated up to 1.2.

@delarosatrevin , any toughts?

azazellochg:
I think three major releases are enough to break backward compatibility.

pconesa:
I thought you were introducing a bug here for old projects causing them not to load.

Looking at it more carefully, I guess _resolution will be always there and mainly:

1.- Previous resolution values stored at _xmipp_resolution, _ctffind_resolution_xxx, etc, will be lost. This might prevent the usage of those ctf outputs

2.- String conversion will lose the resolution for old outputs.

I feel breaking backwards compatibility needs a stronger reason than cleaning the code, which I like a lot to do....I'm not sure what to do...anyone else, reading?

azazellochg:
We need feedback from xmipp team, since I don't know if they use their specific labels. Gctf and ctffind projects always have _resolution now, except very old projects as you said.

delarosatrevin:
I think that the _resolution was introduced a quite long time ago...so I think it is very unlikely that this will affect projects unless very very old. On the other side, a major release is a good moment for cleaning up code, especially here in the core of scipion-em plugin. I'm a bit more in favor of the cleanup, but I don't have any strong arguments.

pconesa:
Ok, let's do it then.

PR #45, opened 2020-01-28
Introducing emlib to centralize binding access

delarosatrevin:
Main changes:
* Now pwem contains **emlib** submodule
* **emlib** depends on an underlying (possibly binary binding) library
* By default xmippLib is used, or the ghost if not installed properly
* **emlib** is grouped mainly in two submodules: **metadata** (MetaData, Row, etc) and **image** (ImageHandler)


pwem/emlib/image/__init__.py
```diff
@@ -24,17 +24,5 @@
...
-
-print("\n >>> WARNING: 'import xmipp' is deprecated for the xmipp binding. "
-      "Please, use 'import xmippLib'")
-stackList = traceback.extract_stack()
-for stackLine in stackList:
-    if 'import xmipp' in stackLine[3] or 'from xmipp import' in stackLine[3]:
-        print("  > Please change 'import xmipp' to 'import xmippLib' in\n"
-              "    File: '%s', line %d.\n" % (stackLine[0], stackLine[1]))
-
-from xmippLib import *
```
dmaluenda: This file was temporary since some time ago the binding was through `import xmipp` and we want to warn developers that the correct way is `import xmippLib`, if they still using that former way.

To be honest, I've never seen that waning triggered, so I will say we can remove this file. If you don't think so, I would keep the warning to be able to remove it in the future.

pwem/emlib/lib.py
```diff
@@ -0,0 +1,30 @@
...
+# *
+# *  All comments concerning this program package may be sent to the
+# *  e-mail address 'scipion@cnb.csic.es'
+# *
+# **************************************************************************
+
+try:
+    from xmippLib import *
+except:
+    from ._libNone import *
```
dmaluenda: I would add a print of the exception, because there are different types of errors and often people report installation issues with the ghost triggered. Then, it will be easier if we have some hint of what is failing.
```
try:
    from xmippLib import *
except Exception as exc:
    from ._libNone import *
    print(exc)
```

pwem/emlib/image/__init__.py
delarosatrevin: Hi @dmaluenda, the module xmip-ghost is gone! I think git diff was confused here because I copy/paste the header file. I agree with you that it has been quite a while with that warning, so if it appears, we should address in the specific case.

delarosatrevin: It is gone now! Thanks for the feeback.

pwem/emlib/lib.py
delarosatrevin: I like the idea if the exception can give some hint about the problem.

pwem/emlib/image/__init__.py
dmaluenda: Sorry you right. Well, the ghost is gone but the _libNone replace it, this is not the point. However, since the file name changed from `.../xmipp.py` to `.../image/__init__.py`, what is gone is the option to call the binding by `import xmipp`. This is my first sugestion so, it's done.

Sorry, I haven't seen the file-name changing in the diff and I simply focused on the content.

Great!

pwem/emlib/lib.py
dmaluenda: Yes, because it's different to get an import error saying 'module not found' (you should install XmippBinaries) or any internal error (we have seen problems when a compiled Xmipp for Ubuntu distros is used in a Centos and the opposite, or other errors regarding libraries colisions).

PR #46, opened 2020-01-31
apply warp in applyTransform

fede-pe:
Pull request to allow the apply transform method to apply warpping to the image.


delarosatrevin:
I have some comments here.

First, we should be careful about minimal API (classes, functions, and parameters), especially in core classes as it is ImageHandler.

Second, regarding the applyTransform method:
* We should have a more explicit documentation of each parameter (now it is missed for warp/wrap)
* Should be applyWarp or applyWrap? I would suggest shorter 'warp' or 'wrap'
* borderValue and borderMean seems redundant, Python is pretty flexible, we could keep a single value, probably 'fillValue' (since this is not applied only in the border) that is usually a float, but can accept special cases such as string 'mean' and then, the mean of the image is used.

delarosatrevin:
@fede-pe , thanks for addressing my suggestions!

pconesa:
@delarosatrevin , @fede-pe , thanks to both.

PR #47, opened 2020-01-31
fix one missing emLib import and make export emdb a more general tool

mmmtnez:
fix one missing emLib import and make export emdb a more general tool

1) emlib import has missing in volume viewer

2) test for exportDB:
   scipion3 tests pwem.tests.protocols.test_protocol_export2DB


pwem/protocols/protocol_export/protocol_export_emdb.py
```diff
@@ -32,19 +32,21 @@
...
+from pwem.convert.atom_struct import fromPDBToCIF, fromCIFTommCIF
 from pwem.emlib.image import ImageHandler
 from pwem.convert import AtomicStructHandler, toCIF
 from pwem.protocols import EMProtocol
 from pwem.objects import FSC
 
 
-class ProtExportEMDB(EMProtocol):
-    """ generates files for volumes and FSCs to submit structures to EMDB
+class ProtExportDataBases(EMProtocol):
```
delarosatrevin: This class name seems duplicated!!!

pwem/protocols/protocol_export/protocol_export_DB.py
```diff
@@ -0,0 +1,233 @@
...
+import pyworkflow.protocol.params as params
+from pyworkflow import VERSION_1_2
+
+from pwem.convert.atom_struct import fromPDBToCIF, fromCIFTommCIF
+from pwem.emlib.image import ImageHandler
+from pwem.protocols import EMProtocol
+from pwem.objects import FSC
+from pyworkflow.utils.path import copyFile
+
+class ProtExportDataBases(EMProtocol):
```
delarosatrevin: This class name (and protocol?) seems duplicated

pwem/protocols/protocol_export/protocol_export_DB.py
```diff
@@ -0,0 +1,233 @@
...
+    VOLUMENAME = 'main_map.mrc'
+    HALFVOLUMENAME = 'half_map_%d.mrc'
+    COORDINATEFILENAME = 'coordinates.cif'
+    MASKNAME = 'mask_%d.mrc'
+    ADDITIONALVOLUMEDIR = "addMaps"
+    ADDITIONALVOLUMENAME = "map_%02d.mrc"
+    MASKDIR = "masks"
+    MASKNAME = "mask_%02d.mrc"
+
+    def __init__(self, **kwargs):
```
delarosatrevin: Empty default constructor is not required.

pwem/protocols/protocol_export/protocol_export_emdb.py
mmmtnez: We have replaced the name of the class name ProtExportEMDB by ProtExportDataBases in order to include other databases in addition to EMDB (as it was suggested by one of the referees of the article doi: 10.1021/acs.jcim.9b01032). We haven't found any other files with the old name ProtExportEMDB.

pwem/protocols/protocol_export/protocol_export_DB.py
mmmtnez: The protocol pwem/protocols/protocol_export/protocol_export_emdb.py doesn't exist anymore, it was replaced by pwem/protocols/protocol_export/protocol_export_DB.py.

pwem/protocols/protocol_export/protocol_export_DB.py
mmmtnez: Removed

pconesa:
basic stuff... @delarosatrevin , which scipion runs your "scipion" command?. 2 or 3?

pconesa:
there is a suspicuis '"path"' (mind the inner " surrounded by '), but not sure why.

delarosatrevin:
@pconesa This is Scipion3, I don't have any Scipion in my path by default.

pconesa:
Thanks, we've just seen the error, @mmmtnez does have maxit and you probably don't and there is a bug in the paths. We'll fix it, thanks  @delarosatrevin

mmmtnez:
OK, thanks  @delarosatrevin, we have already fixed the problem, we have run the test removing maxit and it works as well. We couldn't see that bug because we have maxit and the bug can only appear when you haven't it.

delarosatrevin:
Test are passing now, thanks.

PR #51, opened 2020-02-11
Single config

pconesa:
must be accepted with https://github.com/scipion-em/scipion-pyworkflow/pull/53


azazellochg:
What's this about? Just wondering..

pconesa:
@azazellochg this is to be able to run scipion:
A.-  without any config file (all with reasonable default values)
B.- if you run scipion config (optional due to A) you'll get a single scipion.conf at SCIPION_HOME/config/scipion.conf
C:-Optionally, if you want to have a scipion.conf at ~/.config/scipion it will be taken into account as before, but scipion will not generate it.

pconesa:
You are very welcome to review it, I do not choose to avoid bothering you.

PR #59, opened 2020-02-25
new protocol for orig and sampling modification

rmarabini:
Added new protocol ProtOrigSampling that allows to modify the sampling rate and the origin of coordinates.

Why is this needed? 

1) sampling
  The sampling rate is provided by the microscope setup. This value is usually slightly wrong. For single particle processing, this error does not affect the final results but for model building this value need to be refined otherwise the aminoacids will not fit in the 3D density map. 

2) origin of coordinates: almost any protocol in scipion ignores the value of the 3D map origin and reset it to None. This is catastrophic for model building. The ideal solution would be to redo all protocols and check that the origin is kept. But since I do not wish to review a few hundreds of protocols this protocol is another option.

Test:

 ./scipion3 tests pwem.tests.protocols.test_protocols_assign_orig.TestOriginSampling


pwem/protocols/protocol_origin_sampling_volume.py
```diff
@@ -0,0 +1,173 @@
...
+    # --------------------------- DEFINE param functions ----------------------
+    def _defineParams(self, form):
+        form.addSection(label='Input')
+
+        form.addParam('inVolume', params.PointerParam, label="Input Volume", important=True,
+                      pointerClass='Volume',
+                      help='A new object volume will be created with the assigned sampling rate and origin. No new binary file will be created')
+        form.addParam('copyFiles', params.BooleanParam,
+                      label="copy (true)/link(false) volume:",
+                      help="Option YES:\nA new volume file will be copied"
```
JorMaister: Missing space and a comma between the words "copied" and "otherwise"

pwem/protocols/protocol_origin_sampling_volume.py
```diff
@@ -0,0 +1,173 @@
...
+                      "otherwise a link to the input volume is maded\n"
+                      "default = false", expertLevel=params.LEVEL_ADVANCED,
+                      default=False)
+        form.addParam('setSampling', params.BooleanParam,
+                      label="Set SamplingRate",
+                      help="Option YES:\nA new volume object will be created with "
+                           "the given SamplinRate."
+                           "This SamplinRate will NOT be set in the map file header.\n\n",
```
JorMaister: Missing space between the "." and the word "This".

JorMaister:
Wizards "Pixel size", "x", "y" and "z" displays the values of "x", "y" and "z" with multiple decimal digits.
image

pwem/protocols/protocol_origin_sampling_volume.py
```diff
@@ -0,0 +1,173 @@
...
+                      default=False)
+        form.addParam('samplingRate', params.FloatParam,
+                      condition='setSampling',
+                      label=pwutils.Message.LABEL_SAMP_RATE)
+        form.addParam('setOrigCoord', params.BooleanParam,
+                      label="Set origin of coordinates",
+                      help="Option YES:\nA new volume object will be created with "
+                           "the given ORIGIN of coordinates."
+                           "This ORIGIN will NOT be set in the map file header.\n\n",
```
JorMaister: Missing space between the "." and the word "This"

PR #62, opened 2020-03-03
Devel fixes

pconesa:


pwem/__init__.py
```diff
@@ -79,9 +79,20 @@ class Plugin(pyworkflow.plugin.Plugin):
...
         to the default value.
+
+        If the varName is defined, we will make the value absolute adding SCIPION_HOME
         """
-        cls._defineVar(varName,
-                       os.path.join(Config.EM_ROOT, defaultValue))
+        prefixedDefault = os.path.join(Config.EM_ROOT, defaultValue)
+        varValue = os.environ.get(varName)
+
+        # If en value is not absolute, then we prepend SCIPION_HOME
```
delarosatrevin: I don't see the need for this change.
EM variables can easily be defined without hardcoded 'software/em' and then the proper EM_ROOT will be prefixed.

delarosatrevin: I think we need to clearly define the behaviour of variables (either defined via environment or config file) and avoid unexpected prepending of things. Because this introduce weird behaviour and cases in other contexts. I think this is a moment to make things clear and simple...even if users need to update their old configuration files (which is expected to happen anyway in a major upgrade)

pconesa: then we just need to remove %(EM_ROOT)s and similar cases in the config files, otherwise, we need to do something different. And in any case, just looking at the config it will no be clear what will happen to the variable: some might end up being prefixed with software/em some other not.

What about naming the config files with scipion3.conf and avoiding old configs?

pconesa: I'd say most of the default values (for pyworkflow, em and plugins) should work fine.

PR #67, opened 2020-03-13
Export to DataBases

rmarabini:
Dear all,
we have modified the protocol export2DB to save sampling and origin in the header of each map or mask that the user would like to export.

Test:
scipion3 tests pwem.tests.protocols.test_protocol_export2DB.TestExport2DataBases

### DIFF
--- a/pwem/protocols/protocol_export/protocol_export_DB.py
+++ b/pwem/protocols/protocol_export/protocol_export_DB.py
@@ -30,6 +30,7 @@
 import os
 
 import pyworkflow.protocol.params as params
+from pwem.convert import Ccp4Header
 from pyworkflow import VERSION_1_2
 
 from pwem.convert.atom_struct import fromPDBToCIF, fromCIFTommCIF, \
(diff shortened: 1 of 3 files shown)


pwem/protocols/protocol_export/protocol_export_DB.py
```diff
@@ -133,26 +134,42 @@ def createDirectoryStep(self, dirPath):
...
-        ih = ImageHandler()
-        ih.convert(self.exportVolume.get().getLocation(),
-                   os.path.join(self.dirName, self.VOLUMENAME))
+        inVolFileName = self.exportVolume.get().getFileName()
+        inVol = self.exportVolume.get()
+        outVolFileName = os.path.join(self.dirName, self.VOLUMENAME)
+        shifts = inVol.getOrigin(force=True).getShifts()
+        sampling = inVol.getSamplingRate()
+
+        print ("inVolFileName", inVolFileName)
```
pconesa: I think this print is a debug one?

pwem/protocols/protocol_export/protocol_export_DB.py
```diff
@@ -181,11 +198,19 @@ def exportFSCStep(self):
...
-                       os.path.join(outputDir,
-                                    self.MASKNAME % counter))
+            mask = mask.get()
+            inVolFileName = mask.getFileName()
+            outVolFileName = os.path.join(outputDir,
+                                    self.MASKNAME % counter)
+            shifts = mask.getOrigin(force=True).getShifts()
+            sampling = mask.getSamplingRate()
+
+            ccp4header = Ccp4Header(inVolFileName)
```
pconesa: Are we sure at this point that all masks are mrc? Otherwise we can not use Ccp4Header, or we must convert the mask before this.

pwem/protocols/protocol_export/protocol_export_DB.py
rmarabini: yes, It can be removed (I already did)

pwem/protocols/protocol_export/protocol_export_DB.py
rmarabini: Good point. I do not think we can.  Let me think about it.

    Roberto

rmarabini:
I think the code is OK, see comment in github

    Roberto

pwem/protocols/protocol_export/protocol_export_DB.py
pconesa: ok

pwem/protocols/protocol_export/protocol_export_DB.py
pconesa: ready to merge, then?

rmarabini:
Yes, this is ready to merge

    Roberto

PR #68, opened 2020-03-15
Fix halves import

pconesa:
Hi, this fixes the import of halves, when setting the origin using the right files instead of the main volume.

I've also changed the symmetry of the halves at least to visually see that they are different.


pconesa:
@mmmtnez , will you find time for this PR. Should I "summon" Roberto? ;-)

PR #70, opened 2020-03-20
Jj extract coords assign align

JorMaister:
Tests for extractCoordinates added to xmipp workflow test. 
Protocol extract coordinates fixed. Now it has an advanced option to choose if the shifts should be applied or not.
Protocol alignment assign now have an advanced option used to choose if not to apply the shifts in case they've been applied before (extract coordinates case)


pwem/objects/data.py
```diff
@@ -1579,8 +1579,11 @@ def scale(self, factor):
         m *= factor
         m[3, 3] = 1.
 
-    def scaleShifts(self, factor):
+    def scaleShifts(self, factor, shiftsAppliedBefore=False):
         m = self.getMatrix()
+        if shiftsAppliedBefore:
+            m[0, 3] = m[0, 3] - int(m[0, 3])  # Decimal part of X translation
```
pconesa: I've found this other way to get the decimal part, I like more. But there is no need to change it.

5.55 % 1 --> 0.55

pwem/protocols/protocol_alignment_assign.py
```diff
@@ -47,6 +47,12 @@ def _defineParams(self, form):
         form.addParam('inputAlignment', params.PointerParam, pointerClass='SetOfParticles',
                       label="Input alignments",
                       help='Select the particles with alignment to be apply to the other particles.')
+        form.addParam('shiftsAppliedBefore', params.BooleanParam, default=False,
+                      label='Were particle shifts applied?',
+                      expertLevel=params.LEVEL_ADVANCED,
+                      help='This option should be set to Yes only if the particle shifts have been applied ' +
```
pconesa: I'd simplify a bit the help avoiding mentioning extraction: What about:
Activate this option only if the advanced option "Apply particle shifts" was set to Yes at the extract coordinates protocol.

pwem/protocols/protocol_alignment_assign.py
```diff
@@ -136,3 +142,12 @@ def _validate(self):
...
         return errors
+
+    def _warnings(self):
+        validateMsgs = []
+        if self.shiftsAppliedBefore.get():
+            validateMsgs.append("This option should be set to Yes only if the particle shifts have been applied "
+                                "before.\nIf selected, only the remaining decimal part of the shifts will be applied " +
+                                " for 'x' and 'y'  in the translation component of the transformation matrix. This " +
+                                "will carry longer execution times. Please read the help for further information.")
+        return validateMsgs
```
pconesa: Nice, clear!

pwem/protocols/protocol_extract_coordinates.py
```diff
@@ -388,3 +383,12 @@ def _validate(self):
             errors.append('Input particles do not have alignment information!')
 
         return errors
+
+    def _warnings(self):
+        validateMsgs = []
+        if self.applyShifts.get():
+            validateMsgs.append("Only the integer part of the shifts will be applied in order to avoid " +
```
pconesa: I'd avoid protocol class names:

Only the integer part of the shifts will be applied in order to avoid interpolation. If you are re-extracting particles and want to apply the remaining decimal part of the shifts, ask yes to the advance option 'Were particle shifts applied?' in alignment assign protocol.

pwem/objects/data.py
JorMaister: I like it, it's more elegant and concise.

PR #76, opened 2020-04-02
Modeling workflow test and retry function of atom_struct

mmmtnez:
Dear all,

According to the new output names of scipion-em-chimera protocols, as well as from protocol_coot of scipion-em-ccp4, we have had to modify the assertTrue results in the test test_workflow_modeling.  You can verify that the test is able to run now without failing:

scipion3 tests pwem.tests.workflows.test_workflow_modeling

Running this test we have also observed an additional fail concerning the conversion of format files of atomic structures, specially when we run protocol_emringer, which generates several files and/or directories before failing. We have solve this problem removing all those files/directories before try a new run of the protocol with a different atomic structure file format. Then, we have had to modify the retry method of file pwem/convert/atom_struct.py. To check its functionality the previous test can be run also, as well as the phenix tests: 

scipion3 test --grep phenix --run


rmarabini:
Hi @mmmtnez y @pconesa ,

    I see no major problem with this pull request although I would suggest two small changes::

1)  not to use try: except in order to distinguish between  directories and files. I like better a code as:

===
        if os.path.isfile(file_path) or os.path.islink(file_path):
            os.remove(file_path)
        elif os.path.isdir(file_path):
            shutil.rmtree(file_path)
=====

instead of

===
                            if os.path.exists(os.path.join(cwd, item)):
                                try:
                                    os.remove(os.path.join(cwd, item))
                                except:
                                    shutil.rmtree(os.path.join(cwd, item))
===

2) create a function

  You have repeated  the same code twice. Creating a function would be nice.

pconesa:
will pyworkflow.utils.path.cleanPath() work here?

rmarabini:
Not sure @pconesa , @mmmtnez  correct me if I am wrong. You want to delete the contends  of the directory extra but not the directory extra itself.

 If we use pyworkflow.utils.path.cleanPath()  then the directory extra will be deleted. Of course we can recreate the directory extra but then the inode will be different.

Therefore, pyworkflow.utils.path.cleanPath()  is OK if we recreate the directory and we may assume that no program is acceding files in the OLD extra directory. I guess the assumption is save.

rmarabini:
This pull request is OK with me. Mevertheles I left @pconesa  to has the last decission

mmmtnez:
Done

PR #79, opened 2020-04-12
add new option to volume import. So we can retrieve maps from emdb

rmarabini:
Hi @delarosatrevin y @rsanchezgarc 

 As promised I have added a new option to import volume: Now a 3D map  may be retrieved from EMDB. @rsanchezgarc since you ask for it review it.

test:  
     
     scipion3 tests pwem.tests.protocols.test_protocols_import_volumes
  look to import protoocl  called 'import vol,\n import from EMDB'

### DIFF
--- a/pwem/protocols/protocol_import/volumes.py
+++ b/pwem/protocols/protocol_import/volumes.py
@@ -26,7 +26,10 @@
 # **************************************************************************
 
 
-from os.path import exists, basename, abspath, relpath
+from os.path import exists, basename, abspath, relpath, join
+from os import stat
+from numpy import array
+from numpy.linalg import norm
 
 import pyworkflow.utils as pwutils
 import pyworkflow.protocol.params as params
@@ -35,14 +38,11 @@ import pwem.objects as emobj
 import pwem.convert as emconv
 from pwem import emlib
 from pwem.convert import Ccp4Header
-from pwem.convert.atom_struct import fromPDBToCIF
 
 from .base import ProtImportFiles
 from .images import ProtImportImages
 
 from pyworkflow.utils.path import copyFile
-
-from pwem.convert.atom_struct import AtomicStructHandler
 
 
 class ProtImportVolumes(ProtImportImages):
@@ -208,7 +208,6 @@ class ProtImportVolumes(ProtImportImages):
                         pwutils.createAbsLink(self.half2map.get(),
                                               abspath(self._getVolumeFileName(self.half2map.get())))
 
-
                         vol.setHalfMaps([relpath(self._getVolumeFileName(self.half1map.get())),
                                          relpath(self._getVolumeFileName(self.half2map.get()))
                                          ])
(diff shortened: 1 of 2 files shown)


pwem/protocols/protocol_import/volumes.py
```diff
@@ -128,87 +148,103 @@ def importVolumesStep(self, pattern, samplingRate, setOrigCoord=False):
...
-                    emconv.Ccp4Header.fixFile(self.half2map.get(), newFileName2, origin.getShifts(),
+                origin = emobj.Transform()
+                if setOrigCoord:
+                    origin.setShiftsTuple(self._getOrigCoord())
+                else:
+                    origin.setShifts(x / -2. * samplingRate,
+                                     y / -2. * samplingRate,
+                                     zDim / -2. * samplingRate)
+
+                vol.setOrigin(origin)  # read origin from form
```
rsanchezgarc, line 183: Origin can also be read from EBI rest api when download

```python

 info= json.loads(urllib.request.urlopen("https://www.ebi.ac.uk/pdbe/api/emdb/entry/all/EMD-9812").read())
sampling = info["EMD-9812"][0]["map"]["origin"]

```

pwem/protocols/protocol_import/volumes.py
```diff
@@ -128,87 +148,103 @@ def importVolumesStep(self, pattern, samplingRate, setOrigCoord=False):
...
-
-        for fileName, fileId in self.iterFiles():
-            x, y, z, n = imgh.getDimensions(fileName)
-            if fileName.endswith('.mrc') or fileName.endswith('.map'):
-                fileName += ':mrc'
-                if z == 1 and n != 1:
-                    zDim = n
-                    n = 1
+            # Create a Volume template object
+            vol.setSamplingRate(samplingRate)
```
rsanchezgarc, line 158: When downloaded from EMDB, sampling rate could be taken from rest API , so manual sampling rate in the form should be optional 
You can find the sampling rate of a volume using the code bellow
```python

 info= json.loads(urllib.request.urlopen("https://www.ebi.ac.uk/pdbe/api/emdb/entry/all/EMD-9812").read())
sampling = info["EMD-9812"][0]["map"]["pixel_spacing"]["x"]["value"]

```

rmarabini: Hi @rsanchezgarc ,

I am not sure I agree with you. 

 I think very seldom you want to modify the sampling rate or origin provided by default by the database and if you really want to do so you may use the protocol "pwem - assing Orig and sampling"

Regarding the use of the REST API to get the origin and the sampling rate. Certainly you may get the sampling and the origin as you describe but I do not understand why this is better than the implemented code. Have you find a case in which the data stored in the file header and the data provided by the REST API disagree? If this is the case I will duly  modify the code otherwise I think the implementation is correct.

rsanchezgarc: Hi @rmarabini. I have found a couple of examples in which the header was set to 1 A while the REST API contained the correct information.  But of course, I do not remember which they were. What do you think about doing a double check?.

Regarding the option to modify the origin or sampling rate. in the header I was just thinking on those cases in which you know, for sure, that the numbers are incorrect and you want to amend that.  But I agree that this is not very important (this is why I was thinking on it as an advanced feature)

rmarabini: Hi @rsanchezgarc ,

   I have added: (1) the connection to the EMDB rest api and (2) the check between the information stored in the header and the information retrieved from the API.  I have not added the option to modify the sampling rate and the origin because I can not force it to be an advanced feature (all import volumes share a lot of code). 

 As I had all the code needed I compared the values provided by the API and header for a few hundreds of maps. Certainly there are mistakes in entries with low ID (mainly the sampling in the header is wrong) and everything seems to be OK if you look to the last entries. 

This is the last modification that I implement for this protocol. Either it is aproved as it is or somebody else will need to modify it.

rsanchezgarc: Ok @rmarabini. Seems more than enough for me. Thanks

PR #81, opened 2020-04-21
fix issue https://github.com/scipion-em/scipion-em-relion/issues/122

azazellochg:


pconesa:
maybe not. I guess we want a single alignment assign protocols but at the same time doing thing right for relion...umm.

pconesa:
@delarosatrevin any ideas here? Is not good to have "rln functionality in scipion-em.....but if having it right means changing a lot, I can live with this.

PR #90, opened 2020-05-20
new version of html_report

rmarabini:
New version of html report for facilities. Here we use the databse influxdb to store the data and grafana to display it.

Documentation is available at URL: https://scipion-em.github.io/docs/release-3.0.0/docs/facilities/customize-report

I have not done any test for this. May be It is the rigth time to try mock testing ;-). Since the system requires several password  it is not clear how to implement the test in builbot.

And example is available at URL: https://nolan.cnb.csic.es:3000/d/50ya60eZz/gallery?orgId=1&var-project=test3&refresh=10s but you will need to open a ssh tunnel to acces that address. So run first ssh -L 8888:localhost:3000 scipionweb@nolan.cnb.csic.es and then the URL will be https://localhost:8888/d/50ya60eZz/gallery?var-project=test3

Note, you will need a password to access  grafana (I will send it by email)


azazellochg:
1) I thought the core was frozen for release
2) Why "Here we use X" will be in the core?
3) why not make a plugin for monitors?

delarosatrevin:
I agree that new monitors should go to plugins and not to the core.

rmarabini:
Hi @azazellochg and @delarosatrevin,   (and @pconesa  and @dmaluenda )

   You are really scary sometimes ;-)

>    1. I thought the core was frozen for release

I think the question is:  which is the right branch for the puil request?  To tell you the truth I did not know so I asked Pablo and he suggested devel. From my point of view the correct branch is devel but, if the code is frozen, another branch with the frozen code is needed (may be something called scipion3). 

    2. Why "Here we use X" will be in the core?
    3. why not make a plugin for monitors?

Could you be more specific regarding question 3. In particular how do you think this should be implemented. I have been thinking about it and I do not know how to proceed. At present there are Four monitors plus a summary monitor.  In the monitors I want to introduce a small modification in the getData function, instead of returning all elements  just return the elements with id greater than a certain value. This is a very reasonable and general modification than saves a lot of work to protocol_monitor_summary. Regarding the protocol monitor_summary I like it as it is but I do not want to use the functions available in the file report_html but a different set of functions. That is. I want to modify the way data is exported.

How do you suggest to implement this? The ideal solution would be to keep the monitor protocols were they are and implement  a mechanism that allows users to select between two different ways of exporting the information (I guess different facilities may want to implement there own in the future). That is, I do not think that create another protocol called monitor_system_2 in another repository is a good idea. 

Let us imagine that I place the file report_grafana.py (the equivalent to report_html.py) in another repository. Just the file, I do not want to create a new protocol. I may also create an environment variable with the place where scipion should look for the "exporting agent".

Would this solution be agreeable?

   cheers
   @rmarabini

pconesa:
I think we should start thinking about the impact of the code.  @rmarabini offered himself to implement this and I couldn't say no. If this could go to a plugin or not...we....pwem is a plugin and we can release any version at any time. So probably de discussion should go about: where is the best place for this code.

What about moving everything to a facilities plugins? I'm not in favour of making an effort in decoupling the code and have a "dependency injection" based on an environment variable or something else. I don't see this happening soon and complicates the code.

delarosatrevin:
@rmarabini I don't think we are scary...I fully agree with @azazellochg comments and I think his two main concerns are:
1) Stop adding new code after a frozen point
2) Keep the core (either scipion-pyworkflow or scipion-em) as minimal as possible

I think point 2 is not related to any branch...is just a matter of keep the core really core...adding new code is more code to maintain y and many cases introduce dependencies (and bugs). A core should be as stable as possible. 

@pconesa It is true that pwem is a plugin, but kind of central (or core) one. So my suggestion (not only for this case, but in general) is that we should keep it minimal...since it is a shared asset, that should require more consensus and caring (since many things depends on it). New ideas, e.g protocols, monitors, etc...should be implemented (in my opinion) into existing or new plugins.

rmarabini:
Hi,

Is @pconesa solution agreeable. That is, move the monitors to a plugin oriented toward facility related stuff?.

pconesa:
I guess this can be cancelled since we are creating a facilities plugin.

rmarabini:
Should I delete rm_grafana? I do not need it any longer but I do not know is @dmaluenda  needs it.

dmaluenda:
let me check a couples of things and I will remove it. Thx!

PR #92, opened 2020-05-25
Moving all monitors code to facilities plugin

dmaluenda:
This pull request takes off all code related with monitors in order to migrate it in a [dedicated plugin intended for facilities](https://github.com/scipion-em/scipion-em-facilities).

See below some points I'm not totally sure, please check them:
 - I've considered to also migrate the test_workflow_high_throughput.py
 - I've considered to also migrate the test_workflow_streaming.py
 - May we consider other tests to be also migrate?
 - I think 'nvidia-ml-py3' is a requirement only for monitor GPU usage.

I've checked all remaining tests in 'scipion-em' plugin (having 'scipion-em-facilities' uninstalled) and I can't see any error coming from this migration.

Please, have a look on the 'scipion-em-facilities' plugin and tell me all your concerns. For instance, I doubt for the module's name, the plugin name is 'scipion-em-facilities' but I've chosen 'scipion4facilities' for the module since I don't like just 'facilities'... do you agree? It can be changed....


delarosatrevin:
Yes, I'm fine with moving it to the facilities plugin.

rmarabini:
OK, I am going to tests this installing this branch and plugin scipion-em-facilities

rmarabini:
@dmaluenda, before applying this pull request can you modify a protocol in xmipp which imports ProtMonitor class? The name of the protocol is

xmipp3/protocols/protocol_metaprotocol_discrete_heterogeneity_scheduler.py

I did the following change 

-from pwem.protocols import ProtMonitor
+from pwem import Domain
+ProtMonitor = Domain.importFromPlugin('emfacilities.protocols', 'ProtMonitor', doRaise=True)

Of course, this make xmipp dependent  on emscipionfacilities.  

 y the way, I teste emfacilities and modify the way the secrests.py in handled. Now the file is called scipion.cfg and should be saved in the directory pointed by EMFACILITIES_HOME. I also updatd the documentation.

I think as soon as the change in protocol_metaprotocol_discrete_heterogeneity_scheduler.py is done we are ready

dmaluenda:
No, it's not needed. It's a fake dependency. I've just merged the removing of that fake dependency (https://github.com/I2PC/scipion-em-xmipp/pull/235).

dmaluenda:
Agree with the secrets stuff.

PR #103, opened 2020-06-25
Clean acq

delarosatrevin:
Removed unnecessary field from Acquisition class.


pconesa:
Hi @delarosatrevin I only see changes related to the optics group, but nothing to the Acquisition clean.

delarosatrevin:
well, Acquisition clean I mean removing these attributes from the class. The real clean will be considered later in Relion plugin and then maybe in others.

pconesa:
Ok, then I'm ok with this. I'll aprove it. Feel free to merge it once relion devel branch needs this.

PR #121, opened 2020-07-17
Metadataeditor

pconesa:
This protocol is to generically edit an item in a set. 

We've had 2 cases already for this:

1.- set "amplitude contrast" to a different value of the one specified at import time. It seems (the case is reasonable under certain circumstances)

2.- Set random values to some metadata attributes. This escapes any standard image processing, but was needed to make test and evaluate image processing protocols when data is wrong.

It also passes Scipion DEBUG mode to ShowJ.

### DIFF
--- a/pwem/__init__.py
+++ b/pwem/__init__.py
@@ -38,7 +38,7 @@ import pyworkflow.plugin
 from .constants import *
 from .objects import EMObject
 from .utils import *
-__version__ = '3.0.1'
+__version__ = '3.0.2'
 _logo = "scipion_icon.gif"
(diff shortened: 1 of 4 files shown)


pconesa:
And additional use case:

Modify filenames appending :mrc

This should be right already but we are seeing unexpected headers for subtomograms that scipion undertands like 2d image stacks. Metadata provides a mechanism to fix this cases in case they happen.

cossorzano:
Great

cossorzano:
I would add in the help that numpy is available (and it should be)

PR #126, opened 2020-07-28
centralize chimera visualization

rmarabini:
of a 3D map using another 3D map to set colors (for example the color of the pixels depend on the local resolution)

There are at least 5 places where this very same code has been used. I just deleted from the local viewers an placed in here plus updated it for chimerax


pconesa:
Should the branch be removed?

rmarabini:
yes please

PR #133, opened 2020-08-04
Improve exctract coordinates

pconesa:


pconesa:
Closes https://github.com/scipion-em/scipion-pyworkflow/issues/103

pconesa:
@delarosatrevin will you find time for this? Please.

delarosatrevin:
@pconesa Unfortunately I don't have time to properly review this. I took a quick look. I don't see any comment about micId/micName changes, the issue that is addressed here is related to the summary. Could you comment on the main changes?

pconesa:
Thanks @delarosatrevin  for your time, a good quick look is more than appreciated.

There is no issue related the micId mic name stuff, I just took this branch to quickly fix that issue that is in this PR request but not the main reason for this.

Regarding the micId/micName, one of our betatesters was importing a starfile from relion. This was a set of particles facilitated by a collaborator

He then wanted to "convert them to coordinates" and the extract coordinates was matching the micrographs by the id resulting in a wrong match since the micids from imported particles do not match existing micids in existing sets.

With these changes I propose to try to make the match by micName first....and if not fallback to the micId (existing behaviour) when extracting coordinates.

The rest is a test (mocking some code) to test the matching cases.

delarosatrevin:
@pconesa Could we also change the visibility (EXPERT_LEVEL) of the related options in assing-alignments? The form is quite small anyway. 

Apart from that, everything looks fine. I did some extract tests applying shifts - assigning alignments + reconstruct...and also run all 'extract' tests.

pconesa:
Thank you @delarosatrevin for your thorough review. 
I think I've addressed all suggestions.

delarosatrevin (approved):
Thanks @pconesa

PR #134, opened 2020-08-04
* Fixing the scaleSpline method

fonsecareyna82:


pwem/emlib/image/image_handler.py
```diff
@@ -480,12 +480,20 @@ def scaleFourier(cls, inputFn, outputFn, scaleFactor):
...
+        I.read(inputFn)
+        x, y, z, n = I.getDimensions()
+        if forceVolume:
+            z = max(z, n)
+        if finalDimension is not None:
+            x, y, z = int(finalDimension), int(finalDimension), int(finalDimension)
+        else:
+            x, y, z = int(x * scaleFactor), int(y * scaleFactor), int(z * scaleFactor)
+
+        I.scale(x, y, z, 1 if forceVolume else 0)
```
pconesa, line 496: Can we set n before and avoid here the inline if?

pconesa:
Thank you @fonsecareyna82

PR #135, opened 2020-08-05
allow update class callback in appendFromClasses

pconesa:


pconesa:
@delarosatrevin will you find time for this one too?. Sadly, dependent code is already merged but can be undone if this is not a proper solution.

In summary I needed a way to "clone" a set of 2d classes, but customizing the clone a bit. I changed what is used in the subsets ("literal clone") to pass an optional "callback" to allow customization. The reason for this was in the compare reprojections protocol (the dependent PR): https://github.com/I2PC/scipion-em-xmipp/pull/294

delarosatrevin (approved):
@pconesa I'm OK with the change, seems fine.

pconesa:
Thank you very much!

Issue #136, opened 2020-08-09
Sampling rate is lost on averages set.


pconesa:
https://github.com/scipion-em/scipion-em/commit/ce21d24e803fb39ec4c545035f289d2724a5bde3

azazellochg:
output.copyInfo(inputImages) already supposed to copy sampling rate, why do we need that commit?

pconesa:
That commit is to allow "representatives" (averages) to have a different sampling rate than the images. This was triggered by the fact that CS 2d classification and ab initio, in some cases output downscaled images and the plugin, by mistake was registered them at the same sampling rate as the images.

We are changing this and we want to scale the volumes/averages back to match the images. But also we looked at the option of having different dimensions in the representative and this commit allows it...although accidentally is breaking the sampling rate, which should be easy to fix.

PR #137, opened 2020-08-10
* Fixing an error with the sampling rate creating a set of averages

fonsecareyna82:


fonsecareyna82:
Solution to issue #136

pconesa:
This fixed the cl2d issue. The reason for the bug was that the representative object in relion cl2d and may be other classifiers do not have sampling rate. This PR test for it and if missing the representative sampling rate is taking the sampling rate from the class, which in this cases should match.

azazellochg:
Great! Thank you both!

PR #139, opened 2020-08-18
Jj cryo care prep

JorMaister:


pconesa:
Hi @JorMaister , I'm not sure if this is the right way....I'm of until Monday. Can this wait?

The main reason is not to overload this protocol that I think is already overloaded with more non obvious functionality.


Also this protocol is base for many of them we need to be cautious.

Give me some time to discuss this with you and may be others.

JorMaister:
HI @pconesa ,
We're not in a hurry with it, so we can wait and see this on Monday.

pwem/protocols/protocol_align_movies.py
```diff
@@ -105,6 +105,16 @@ def _defineAlignmentParams(self, form):
                       label="Save aligned micrograph",
                       expertLevel=pwcts.LEVEL_ADVANCED)
 
+        form.addHidden('tomographySplitFrames', params.BooleanParam,
```
delarosatrevin, line 108: I think tomography stuff should not be at this level...if base align movies protocol is needed, maybe a new class or a sub-class of this one.

pwem/protocols/protocol_align_movies.py
```diff
@@ -105,6 +105,16 @@ def _defineAlignmentParams(self, form):
...
+                       default=getattr(self, 'tomographySplitFrames', False))
+        form.addParam('splitEvenOdd', params.BooleanParam,
+                      default=False,
+                      label='Split & sum odd/even frames?',
+                      condition='tomographySplitFrames',
+                      expertLevel=pwcts.LEVEL_ADVANCED,
+                      help='(Used for cryoCARE denoising data preparation). If set to Yes, 2 additional tilt series '
```
delarosatrevin, line 115: Even more, if it is only for cryoCARE, it should be analyzed where it is the best place for this option.

pconesa:
We are sending this, at least to draft trying to find a better approach not involving this package.

pwem/protocols/protocol_align_movies.py
pconesa: Theoretically, any denoiser or maybe any other method might use this even / odd data. So I'd say is not cryoCARE specific but needed by cryoCARE.

delarosatrevin: I guess this will depend on the specific algorithm used by the denoiser, I don't know about that, but probably not all of them will require the same input.

pconesa: I'm not an expert either, I just see that splitting even and odd has been done in the past by xmipp team for something else and is the same kind of splitting. But time will tell.

PR #142, opened 2020-09-08
Bug fix, for negative numbers the module does not remove the integer …

cossorzano:
…part

This fixes the test xmipp3.tests.test_protocols_xmipp_2d.TestAlignmentAssign

### DIFF
--- a/pwem/objects/data.py
+++ b/pwem/objects/data.py
@@ -1577,8 +1577,8 @@ class Transform(EMObject):
     def scaleShifts(self, factor, shiftsAppliedBefore=False):
         m = self.getMatrix()
         if shiftsAppliedBefore:
-            m[0, 3] = m[0, 3] % 1  # Decimal part of X translation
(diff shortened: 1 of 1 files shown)


pwem/objects/data.py
```diff
@@ -1577,8 +1577,8 @@ def scale(self, factor):
     def scaleShifts(self, factor, shiftsAppliedBefore=False):
         m = self.getMatrix()
         if shiftsAppliedBefore:
-            m[0, 3] = m[0, 3] % 1  # Decimal part of X translation
-            m[1, 3] = m[1, 3] % 1  # Decimal part of Y translation
+            m[0, 3] = m[0, 3] - int(m[0,3])  # Decimal part of X translation
```
delarosatrevin, line 1580: What about:
```python
m[0, 3] -= int(m[0, 3])
m[1, 3] -=  int(m[1, 3])
```

cossorzano: OK

pconesa (approved):
Nice.

PR #143, opened 2020-09-09
Expect None in some values when converting to str a ctfModel. There a…

pconesa:
…re already cases where resolution a fit quality are missing from ctf refinement protocols. It might be a bug, but this tolerates future bugs.

### DIFF
--- a/pwem/objects/data.py
+++ b/pwem/objects/data.py
@@ -138,12 +138,12 @@ class CTFModel(EMObject):
         phaseShift = self.getPhaseShift() if self.hasPhaseShift() else 0
         ctfStr = "defocus(U,V,ast,psh,res,fit) = " \
                  "(%0.2f,%0.2f,%0.2f,%0.2f,%0.2f,%0.2f)" % \
-                 (self._defocusU.get(),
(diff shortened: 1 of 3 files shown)


azazellochg:
@pconesa what are "ctf refinement protocols"?

pconesa:
Need to check that but I believe one of the particle CTF refinement one. Thera was a gCTF and I believe Xmipp has one too.

pconesa:
Done!

pwem/objects/data.py
```diff
@@ -136,14 +136,18 @@ def __init__(self, **kwargs):
 
     def __str__(self):
         phaseShift = self.getPhaseShift() if self.hasPhaseShift() else 0
```
delarosatrevin, line 138: I think now you can treat phaseShift the same as the other ones.

delarosatrevin (approved):
Nice @pconesa . Just a minor comment left.

pwem/objects/data.py
pconesa: Yep, done.

PR #144, opened 2020-09-11
Do not write axis file if not to be shown

pconesa:
Do not show axis for local resolution viewers

Tweaks to not show axis in local resolution viewers by default.

image

### DIFF
--- a/pwem/viewers/viewer_chimera.py
+++ b/pwem/viewers/viewer_chimera.py
@@ -396,22 +396,27 @@ def mapVolsWithColorkey(displayVolFileName,
     fhCmd.write("from PyQt5.QtGui import QFont\n")
     fhCmd.write("run(session, 'set bgColor %s')\n" % bgColorImage)
 
-    bildFileName = scriptFile.replace(".py",".bild")
-    Chimera.createCoordinateAxisFile(voldim[0],
-                                     bildFileName=bildFileName,
-                                     sampling=sampling)
-    # axis
     chimeraVolId = 1
+
     if showAxis:
+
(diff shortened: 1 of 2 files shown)


rmarabini:
Hi @pcosena,

      There are a couple of lines I do not understand:

  ```        if showAxis:
            x = -voldim[0] * sampling // 2
            y = -voldim[1] * sampling // 2
            z = -voldim[2] * sampling // 2
        else:
            x = y = z = 0

Why should be origin placed at (0,0,0) (a corner) instead of at the volume center?

I think you wan to change this line

         fhCmd.write("run(session, 'cofr 0,0,0')\n")  # set center of coordinates

so instead of (0,0,0) is xdim/2,ydim/2, zdim/2) and  leave the origin in the default place.

COFR places de center of rotation.

     Roberto

pconesa:
That is the way I made it work, when axis file is absent. I tried to keep
the code line but didn't work.

Maybe we can find some time to play with these values and find  a better
solution.

rmarabini:
Hi @pconesa 

  Can we meet and discuss this change? Of course is up to the user if he wants to display the axis but I think the real problem here is a confusion between what is the origin and what is the center of rotation. My feeling is that the volume origin should not be modified for display purposes but there is no problem changing the center of rotation.

  For example, imagine I want to apply a symmetry o merge with other data. If the origin has been modified these two operations will fail but if the center of rotation changes there is no problem

rmarabini:
such a nice commit :-)

rmarabini:
After  videoconferencing we reach an agreement

PR #148, opened 2020-09-20
Rm assign angles

rmarabini:
Hi @pconesa, 

    This is my last PR for scipion3. It modifies the alignment matrices so that the handness of the reconstruction is modified.

   I also added a test for the protocol assign_angles which had none.

TEST:  scipion3 tests pwem.tests.protocols.test_protocol_alignment_assign

### DIFF
--- a/pwem/protocols/protocol_alignment_assign.py
+++ b/pwem/protocols/protocol_alignment_assign.py
@@ -23,7 +23,7 @@
 # *  e-mail address 'scipion@cnb.csic.es'
 # *
 # **************************************************************************
-
+import time
 
 import pyworkflow.protocol.params as params
 from pyworkflow.object import Integer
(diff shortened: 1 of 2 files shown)


delarosatrevin:
I think it should be cleaner to have a separate protocol to invert the hand...and it can be used either after assign - alignment or after any protocol that produce aligned particles.  In that way the assign - alignment will remain unchanged and with simpler logic...and the createOuput much simpler without the need to validate that input particles and alignments are the same, that makes no sense apart from invert hand.

rmarabini:
This claim of more protocol decrease complexity may be made virtually for  any protocol in scipion. 

In this case both ways make sense, Two protocols makes the coding of each one slightly simpler. A single protocol avoid a lot of code duplication. You can always make a base class and then inherent from there but then I do not think both protocols are simpler.

delarosatrevin:
Yes, I agree that unnecessary inheritance should be avoided. In this particular case, I think that inverting hand to the same set introduces an awkward combination of using the same input twice and introduces a weird case in the createOutput...a separate protocol that just invert the hand should not have a lot of code duplication since it only need to basically implement the copyItems in createOutput step and update the matrix for each item.

pconesa:
Thank you @rmarabini for your work and @delarosatrevin for giving your feedback.

So....I'm happy with whatever you both agree on.

rmarabini:
I think I will follow @delarosatrevin
<https://github.com/delarosatrevin>


  cheers

delarosatrevin:
Thanks @rmarabini . I think in this case will be better this way, although you're right that in other cases it will be better to avoid code duplication. I'm sorry for the extra work, let me know if I can be of any help... and thanks a lot for adding tests for this protocol!!!

PR #150, opened 2020-09-25
Rm atom struct

rmarabini:
add renumber and rename functions to atom_struct.py which allows to shift the residues  position by an offset and change the chain name

Also a have modified the format used by invert_hand protocol to print the progressbar 
since scipion uses variable fonts so the bar size changes in each print
because   the "space" width is different from the "=" width

### DIFF
--- a/pwem/convert/atom_struct.py
+++ b/pwem/convert/atom_struct.py
@@ -689,13 +689,13 @@ class AtomicStructHandler:
         io.set_structure(self.structure)
         io.save(filename, sel)
 
-    def reNameChain(self, chainID, newChainName, modelID='0',
+    def renameChain(self, chainID, newChainName, modelID='0',
                     filename="output.mmcif"):
         self.structure[modelID][chainID].id = newChainName
         self.write(filename)
 
 
-    def reNumberChain(self, chainID, offset=0, modelID='0',
+    def renumberChain(self, chainID, offset=0, modelID='0',
                     filename="output.mmcif"):
         # get chain object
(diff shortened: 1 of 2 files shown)


pwem/convert/atom_struct.py
```diff
@@ -689,6 +689,34 @@ def extractChain(self, chainID, start=0, end=-1,
         io.set_structure(self.structure)
         io.save(filename, sel)
 
+    def reNameChain(self, chainID, newChainName, modelID='0',
```
delarosatrevin, line 692: Weird camelCase...what about `renameChain` ?

pwem/convert/atom_struct.py
```diff
@@ -689,6 +689,34 @@ def extractChain(self, chainID, start=0, end=-1,
...
+                    filename="output.mmcif"):
+        self.structure[modelID][chainID].id = newChainName
+        self.write(filename)
+
+
+    def reNumberChain(self, chainID, offset=0, modelID='0',
```
delarosatrevin, line 698: `renumberChain` ?

pwem/convert/atom_struct.py
rmarabini: done

pwem/convert/atom_struct.py
rmarabini: done

rmarabini:
refactorization done

delarosatrevin (approved):

pconesa (approved):

PR #153, opened 2020-09-30
Call Set method (appendFromSet) if exists to delegate the implementation

pconesa:
This is basically the if but I've , additionally, renamed some variables to reflect better what they are.


pconesa:
Replaced by https://github.com/scipion-em/scipion-em/pull/221

PR #155, opened 2020-10-01
Replaced create by createCopy function, that also can copyInfo/copyItems

delarosatrevin:


pwem/objects/data.py
```diff
@@ -931,6 +932,39 @@ def copyItems(self, otherSet,
...
+        fn = prefix or self.getClassName().lower().replace('setof', '')
+
+        if suffix:
+            fn += '_%s' % suffix
+
+        fn += '.%s' % (ext or pwutils.getExt(self.getFileName()))
+
+        setPath = os.path.join(outputPath, fn)
+        pwutils.cleanPath(setPath)
+
```
pconesa, line 959: The mapper part is missing:

        import pyworkflow.mapper as pwmapper
        pwmapper.SqliteDb.closeConnection(setFn)

To be hones I do not remember why is there, but I guess it was needed.

pconesa: May be @fonsecareyna82 remembers in which sceneario we had to add the closeConnection line.

delarosatrevin: I added this in the Mapper itself...look here https://github.com/scipion-em/scipion-pyworkflow/pull/150

pconesa: I see... thanks.

Could we keep, for now the "create" name, or have an additionally create one. Mostly in tomo plugins they are using the create() avoiving the traditional self._createSetOfXXX that was the problem for this and for the subset.

delarosatrevin: "create" is quite ambiguous, I think that we are still on time to refactor this. I would prefer to have a more explicit name for this operation.

delarosatrevin: By the way,  I think the original case  in  TS, we should use another name for copyItems (maybe copyTs or something else), since the copyItems change  the base signature from EMSet.

pconesa: I was aiming for a common signature, so subsets, join all set operation protocols can create, copy regardless the type.

delarosatrevin: Yes, the createCopy have a common signature and copyItems should have it as well...I'm just pointing out, that maybe in TiltSeries, the current copyItems should have a different  name, since it does not respect  the  base signature  inherited from EMSet. 

PS:  I think  I added that myself :(

pwem/objects/data.py
```diff
@@ -870,8 +872,7 @@ def __init__(self, filename=None, pseudoatoms=False, **kwargs):
 class EMSet(Set, EMObject):
```
pconesa, line 872: What about declaring here:
USE_CREATE_COPY_FOR_SUBSET = False

And avoiding hasAttr in the protocol batch? just  `if modifiedSet.USE_CREATE_COPY_FOR_SUBSET` ?

pwem/protocols/protocol_batch.py
```diff
@@ -185,7 +182,7 @@ def _createSubSetFromImages(self, inputImages,
             createFunc = getattr(self, '_create' + className)
             output = createFunc()
         except Exception:
-            output = inputImages.create(self._getPath())
+            output = inputImages.createCopy(self._getPath())
```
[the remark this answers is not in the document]
delarosatrevin: You are calling it from an instance, so I think it is better the createCopy, but it does not make a big difference, the createCopy can avoid a few extra lines.

pwem/objects/data.py
delarosatrevin: I don't like it since it is something that should be deprecated, and it is not expected to be used in other places (remember that subset logic does not necessarily be the one with the gui/batch one). I would keep the getattr with default False there...and keep the EMSet class more clean.

pconesa:
This is replaced by https://github.com/scipion-em/scipion-em/pull/221/

PR #167, opened 2020-11-03
Aj union set volumes

ajimoreno:
The protocol "pwem - join sets" has been modified to take several Volumes as input and create a SetOfVolumes as output.

### DIFF
--- a/pwem/protocols/protocol_sets.py
+++ b/pwem/protocols/protocol_sets.py
@@ -38,6 +38,7 @@ import pyworkflow.object as pwobj
 
 import pwem.objects as emobj
 from pwem.protocols import EMProtocol
+from pwem.objects import Volume
 
 
 class ProtSets(EMProtocol):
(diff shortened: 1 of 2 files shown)


pwem/protocols/protocol_sets.py
```diff
@@ -139,13 +140,19 @@ def createOutputStep(self):
 
         # Read ClassName and create the corresponding EMSet (SetOfParticles...)
         try:
-            outputSetFunction = getattr(self, "_create%s" % set1.getClassName())
+            if str(set1.getClassName()) is not 'Volume':
```
pconesa, line 143: I now 'Volume' is in many places, but we should try to improve this having a single class name source.

I'd suggest to replace 'Volume' by  `Volume.__name__`

pwem/protocols/protocol_sets.py
```diff
@@ -139,13 +140,19 @@ def createOutputStep(self):
...
+                outputSetFunction = getattr(self, "_create%s" % set1.getClassName())
+            else:
+                outputSetFunction = getattr(self, "_createSetOfVolumes")
```
pconesa, line 146: Can't this be replaced by "self._createSetOfVolumes" ?

pconesa (changes requested):
Thanks @ajimoreno 
I've got some minor changes to do.

pwem/protocols/protocol_sets.py
ajimoreno: Done

pwem/protocols/protocol_sets.py
ajimoreno: Done

pconesa: Thanks @ajimoreno I've added a test for this case

pconesa (approved):

PR #183, opened 2021-01-08
create DONE folder in extra

azazellochg:
Ctf and picking protocols do have a DONE folder in extra, but movie alignment protocols seem to dump txt files directly into extra. This should fix this.

### DIFF
--- a/pwem/protocols/protocol_movies.py
+++ b/pwem/protocols/protocol_movies.py
@@ -116,7 +116,7 @@ class ProtProcessMovies(ProtPreprocessMicrographs):
 
     # STEP to convert correction images if apply
     def _convertInputStep(self):
-
+        pwutils.makePath(self._getExtraPath('DONE'))
         movs = self.inputMovies.get()
 
         # Convert gain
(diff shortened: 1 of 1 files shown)


pwem/protocols/protocol_movies.py
```diff
@@ -116,7 +116,7 @@ def _insertAllSteps(self):
 
     # STEP to convert correction images if apply
     def _convertInputStep(self):
-
+        pwutils.makePath(self._getExtraPath('DONE'))
```
azazellochg, line 119: My only concern is if this is a right place to do this.

pconesa (approved):

pwem/protocols/protocol_movies.py
pconesa: Yeah! I understood the reason. Another way this could be done, freeing up the "when" is something like:

def getDONEFile(self, *paths):

 if self._doneFolder is None:
    self._doneFolder = self._getExtra("DONE")
    if not os.path.exists(self._doneFolder):
        pwutils.makePath(self._doneFolder)

 return os.path.join(self._doneFolder, *paths)

This way, "DONE" is only in one place. And creation happens only once the first time is called. In exchange, there is always an "if to evaluate".

pconesa: Hopefully we will avoid all these done files soon.

PR #202, opened 2021-03-16
Optional Y flipping of the output micrographs for managing the tiff f…

DaniDelHoyo:
…ormat (used by motioncorr2 and unblur (cistem))

### DIFF
--- a/pwem/protocols/protocol_align_movies.py
+++ b/pwem/protocols/protocol_align_movies.py
@@ -232,6 +232,11 @@ class ProtAlignMovies(ProtProcessMovies):
 
             for movie in newDone:
                 mic = micSet.ITEM_TYPE()
+                if hasattr(self, "flipY") and self.flipY:
+                    print('Flipping Y output micrograph')
+                    micPath = self._getExtraPath(getOutputMicName(movie))
+                    newMcFn = self.flipYGain(micPath, micPath)
+                    movie.setMicName(newMcFn)
                 mic.copyObjId(movie)
                 mic.setMicName(movie.getMicName())
                 # The subclass protocol is responsible of generating the output


azazellochg:
This will not be implemented in mc2/cistem plugins. 

 In an MRC file, the origin is bottom-left, +X to the right, +Y to the top. Tiff has the origin at top-left, +X to the right, +Y to the bottom. So, the origin and the direction of the Y axis are the opposite between MRC and TIFF. IMOD, EMAN2, SerialEM, MotionCor2, Warp and Cryosparc flip the Y axis whenever they read or write a TIFF file. I don't know what xmipp does.

Now, the problem happens when using dm4 gain reference together with this tiff movie. I'm using example from scipion test dataset. Here is a simple project, where in case (1) gain is dm4, in case (2) gain was converted by dm2mrc, or (3) by e2proc2d to mrc.

image

 In case (1) any following motion correction (mc2 / relion / unblur) will output a bad mic:
Screenshot_20210316_194538

In cases 2 and 3, all mc programs will output good result:
Screenshot_20210316_194837

In all 9 cases the output mics have same orientation, but the gain is applied wrongly in the case 1. Case 1 can be fixed by flipping the gain image upside down inside mc protocol. From this I conclude that xmipp does not do that flip when converting dm4 to mrc, compared to imod / eman2. Tif movies flipping must be a separate issue.

delarosatrevin:
@azazellochg I think the origins of the coordinates and increasing direction of the axis is more a visualization thing, not a data representation of a given format. Anyway,  your tests seem to prove that there is an issue with Xmipp conversion of dm4 or tiff. 

Still, I can't see the need for the proposed changes (here and in the mc plugins)...since each program has options to manipulate the input gain image (flip, rotate,  etc). @DaniDelHoyo, can you comment more on the motivation of your proposed changes? And why is required more code to deal with something that seems already covered?

DaniDelHoyo:
Hello Grigory, I have checked several datasets with different combinations of formats movies-gain (tif-mrc, tif-dm4, mrcs-mrc), including the one you are talking about. Also, I have checked unblur, mc2 and flexalign (still have to check relion). 
I addressed at the same time the dm4 gain and the tiff movies flipping and I have assumed that the movie gain protocol has been used to correct the gain orientation (in your example it is already correct). 

With the last changes (also 1 change in the movie_align protocol in scipion-em), the outputs for these 3 protocols are the same:
- Micrographs are flipped if movie format is tiff, so the orientation is maintained for all of them.
- The gain is always converted to mrc and it's flipped if movies are tif, which results in a proper gain correction. 

I have double checked these changes and I think they work fine, but if you think we should address these problems in a different way I am also okay with that.

azazellochg:
Thank you for the explanation of the suggested changes. Could you explain why none of the existing cryo-em software has this problem and only xmipp does? Is it because xmipp tiff reader does not flip the axis? If it has not been doing so for the last 20 years, I'd rather leave it "as is" not to break any compatibility. Also, in the suggested changes you seem to flip (I might be wrong) movies, gain and then mics again. This will result in unnecessary slowdown of any processing especially with changes done in python and not in xmipp core (why?). To me the simplest solution is available already in the existing code which is only flip the gain once. 
If you have checked that gain only requires flipping when working with tif movies, why I do not see it in tif/mrc case? Eman/imod doesnt know anything about tif movies when I convert dm4 to mrc and the mc works.

DaniDelHoyo:
Hello again, let's go step by step.
- Could you explain why none of the existing cryo-em software has this problem and only xmipp does? Is it because xmipp tiff reader does not flip the axis?:
About the xmipp reader and the other cryo-em software, I'm quite a newbie in this field so I could not tell. I just know that when the gain image is oriented with respect to tiff movies, only flexAlign (xmipp) would give a correct gain correction. Maybe it was because the orientation is also done with xmipp and they share the "flipping" state. Yet, in my personal opinion, this gain orientation step is quite useful since I think there is no convention about how to store the gain, talking about its orientation, in the databases so it is a way to standardize. But again, this is my opinion.

- Also, in the suggested changes you seem to flip (I might be wrong) movies, gain and then mics again. This will result in unnecessary slowdown of any processing especially with changes done in python and not in xmipp core (why?):
About all the flips in my changes, I do not include movie flipping precisely because I guess it would result in a slowdown as you say. As I understand it, tiff movies are by default flipped in all these cry-em software (because of the different origin), so the thing I do is also flipping the gain, so it still matches in the correction process. Then, after the correction, as the micrographs are mrc I want the to recover the original orientation, so I flip them again. Without this, the micrographs would be flipped with respect to those obtained with flexalign, which I supposed was the correct orientation, I might be wrong, but at least for now I have make them all equal. I don't think these flips are too expensive in time, but if you think it must be done differently or in xmippCore, again I'm a newbie so I will do as you prefer.

- If you have checked that gain only requires flipping when working with tif movies, why I do not see it in tif/mrc case? Eman/imod doesnt know anything about tif movies when I convert dm4 to mrc and the mc works.:
I am not sure about this, I just can think maybe the gain was already flipped with respect to the movies so it worked fine, and for me that I first "corrected" the orientation the problem appeared. But I guess you are talking about several examples, so I am not sure. I just can tell that using the orientation correction and then these last changes, it seems to work fine.

-

azazellochg:
I already regret I raised this issue..

I'm not using xmipp, but all other mc programs have no problems with any orientations once the gain is converted from dm4 to mrc. Of course this also depends on how acquisition was done. E.g., SerialEM never rotates the original gain reference coming from GMS, but the movies can be rotated / flipped depending on the camera settings on the scope. Rotation/flip flag is saved in the tif header in the ImageDescription tag. I suggest to wait for @cossorzano who can enlighten us what xmipp does when reading tif files compared to other programs.

azazellochg:
I think my conclusion would be not to change anything. If the gain needs to be flipped in scipion then flip it in the protocol. All software are different and we can't have a single solution. The only thing to be worried is mrc format pixel order / display to be handled properly :)

cossorzano:
Dear all,

I cannot tell exactly what Xmipp does with TIFF files (as I did not do 
that part), but Daniel has done all the important tests. I would say 
that there are two important considerations:

- It is important to keep the possibility inside each protocol to 
flip/rotate the gain if required, as it is already implemented. This 
manual intervention may be needed if the automatic corrections fail.

- Scipion should work out-of-the-box in as many different scenarios as 
possible and identified. The idea is that the same workflow can work in 
many different facilities without the need of the operator to figure out 
which is the right combination of flipping/rotation required for his/her 
setup. This implies, indirectly, that there is a common behavior of the 
different plugins doing similar tasks, and in particular movie alignment 
(in the same way as the decision of the origin of the X and Y 
coordinates of particles in micrographs within Scipion is defined by 
Scipion and it is not left to the user or the program below the 
protocol). Thanks to this common definition, it is easy to substitute 
one aligner by another.

This latter has been the work of Daniel in the last weeks and he has 
performed all the tests he has imagined (as always, there could be 
situations we did not test and we will have to fix, but what is known is 
that the current situation, not doing anything, is not working in all 
scenarios without manual intervention). I would suggest that the 
solution should be the one that implies the minimum amount of changes 
(in whichever plugin) so that all movie aligners behave in the same way 
for all the tests Daniel has defined. Would that be fine with everybody?

Cheers, Carlos Oscar

delarosatrevin:
Dear Carlos, 

I think the main point of @azazellochg and me here is that the flipping behavior can already be achieved with the existing options in the mc protocols, so we don't see the point to add additional code to maintain. I think it is impossible to avoid manual intervention with such a variety of combinations of formats/acquisition software/processing software. I haven't considered (or checked) all scenarios...but I will be more confident if we don't add another source of confusion in Scipion if we do our own flipping and this will require manual intervention anyway for some of the programs. 

I don't really like the idea of doing extra data modification that the underlying program is not doing. This goes against one of the Scipion core principles that we should stick as close to the behavior of the original program as possible. 

If you think you have found an empirical solution to figure out the flipping parameters for some combinations that avoid manual intervention, this fits more into a wizard, metaprotocol that sets the mc params or into a protocol that does the flipping explicitly. 

Cheers,
Jose Miguel

cossorzano:
But this proposal precludes automatic workflow setup without manual 
intervention, does it not?

El 18/03/2021 a las 14:00, Jose Miguel de la Rosa Trevin escribi??:
>
> Dear Carlos,
>
> I think the main point of @azazellochg 
> <https://github.com/azazellochg> and me here is that the flipping 
> behavior can already be achieved with the existing options in the mc 
> protocols, so we don't see the point to add additional code to 
> maintain. I think it is impossible to avoid manual intervention with 
> such a variety of combinations of formats/acquisition 
> software/processing software. I haven't considered (or checked) all 
> scenarios...but I will be more confident if we don't add another 
> source of confusion in Scipion if we do our own flipping and this will 
> require manual intervention anyway for some of the programs.
>
> I don't really like the idea of doing extra data modification that the 
> underlying program is not doing. This goes against one of the Scipion 
> core principles that we should stick as close to the behavior of the 
> original program as possible.
>
Beside this principle, another principle in Scipion is defining internal 
conventions that allow the interaction of different programs. That is 
where the convertInput and createOutput steps fit in.
>
> If you think you have found an empirical solution to figure out the 
> flipping parameters for some combinations that avoid manual 
> intervention, this fits more into a wizard, metaprotocol that sets the 
> mc params or into a protocol that does the flipping explicitly.
>
This is what we have done for the gain (a protocol that changes its 
orientation), but the problem with TIFF is different. It is that the 
underlying programs seem to treat differently MRC files from TIFF files 
(probably because the readers are different).

So, what is next? Closing the pull request or accepting it?

Cheers, Carlos Oscar

pconesa:
Hi all!.

I'd like to thank everyone here for contributing. Especially Daniel for proposing a solution and assessing the current scenario. Also Grigory for taking time to test current behaviour in and out of Scipion with different methods and formats. 

I'd also like to suggest everyone to take some time to listen each other's point of view before making any decision. I think this will facilitate understanding each other and reaching a better solution that I believe we all want. Don't take me wrong, we need to keep the critical review at the same level, is the right thing to do.

Now, I've been talking to Daniel about this and we have decided to redo it differently, As it was, it was impacting many protocols.
I'll try to summarize what Daniel has thought to simplify this:

0.  Assumption: Gain orientation and movie orientation could be a unpredictable and depends very much on controlling the acquisition info.
1. No change will be required for MC, unblur ,  etc plugins
2. Movie gain protocol will output the gain in the most compatible way with default values of all movie alignment protocols. This   is mostly "gain aligned with movies orientation" except for the tiff movies, where the gain will be flipped on Y. This way most protocols will run fine with default values except flexalign (see 3)
3. Flex align will internally "flip the gain" in case movie are in tiff format. By the way, I do not see a user option to flip the gain in the flexalign form in case he/she wants.

Am I missing anything? Does this fit with your understanding of this issue.

Now, with this, all stays the same, plus the optional movie gain protocol will try to provide a solution for the gain orientation when the user does not know it and is hard to know.

I'd also like to point out that, if I haven't missed anything...when using tiff files, the orientations of the micrographs will be different between flexalign and the rest. This means, that you can not reuse coordinates picked on mics outputted by MC2 on flexalign mics without a  "flipping action", mics or coordinates, which I do not think is possible now.

delarosatrevin:
@cossorzano, as @pconesa already proposed, there different options that don't preclude automatic workflow setup/run. 
Regarding tiff and orientation...if program A flips Y...and program B doesn't, what will Scipion do when importing tiff images? There is not correct change and then no way to have a comment convention. 

@pconesa I think the proposed solution is very good in the sense that achieves the same goals and minimizes the impact of required changes.

cossorzano:
Dear all,

El 18/03/2021 a las 14:32, Pablo Conesa escribió:
>
> Hi all!.
>
> I'd like to thank everyone here for contributing. Especially Daniel 
> for proposing a solution and assessing the current scenario. Also 
> Grigory for taking time to test current behaviour in and out of 
> Scipion with different methods and formats.
>
> I'd also like to suggest everyone to take some time to listen each 
> other's point of view before making any decision. I think this will 
> facilitate understanding each other and reaching a better solution 
> that I believe we all want. Don't take me wrong, we need to keep the 
> critical review at the same level, is the right thing to do.
>
> Now, I've been talking to Daniel about this and we have decided to 
> redo it differently, As it was, it was impacting many protocols.
> I'll try to summarize what Daniel has thought to simplify this:
>
>  0. Assumption: Gain orientation and movie orientation could be a
>     unpredictable and depends very much on controlling the acquisition
>     info.
>  1. No change will be required for MC, unblur , etc plugins
>  2. Movie gain protocol will output the gain in the most compatible
>     way with default values of all movie alignment protocols. This is
>     mostly "gain aligned with movies orientation" except for the tiff
>     movies, where the gain will be flipped on Y. This way most
>     protocols will run fine with default values except flexalign (see 3)
>  3. Flex align will internally "flip the gain" in case movie are in
>     tiff format. By the way, I do not see a user option to flip the
>     gain in the flexalign form in case he/she wants.
>
That's correct, we don't have this possibility, and eventually it may be 
needed inside the protocol. What we did was to externalize it to the 
movie gain protocol.

In general, this is also what Daniel and myself talked yesterday, 
finding the minimum amount of changes.

> Am I missing anything? Does this fit with your understanding of this 
> issue.
>
> Now, with this, all stays the same, plus the optional movie gain 
> protocol will try to provide a solution for the gain orientation when 
> the user does not know it and is hard to know.
>
> I'd also like to point out that, if I haven't missed anything...when 
> using tiff files, the orientations of the micrographs will be 
> different between flexalign and the rest. This means, that you can not 
> reuse coordinates picked on mics outputted by MC2 on flexalign mics 
> without a "flipping action", mics or coordinates, which I do not think 
> is possible now.
>
Yes, and this is a problem we have to fix inside Flexalign so that all 
protocols solving the same problem (movie alignment) are interchangeable 
for all input file formats. But, this is calling for a convention, or at 
least a test, inside Scipion that any movie alignment protocol with 
these movies produce an output in this orientation. Otherwise, we may 
encounter the same problem later with future alignments.

Cheers, Carlos Oscar

azazellochg:
I'd say if you want to make a convention then it should be the one followed by most packages. If only cryoSPARC and xmipp do not flip tifs when reading/writing, then the convention would be to flip by default. This means that cryoSPARC and xmipp plugins have to flip it in their code to follow this new Scipion "convention".

azazellochg:
And then you will have cryosparc users complaining that the results in cryosparc inside and outside Scipion are different.

DaniDelHoyo:
Dear all,

as Pablo and Carlos Oscar said, I have made new different code which try to avoid changing the mc2, cistem and relion plugins. Instead, the movie gain protocol outputs the movies with the associated gain as it is expected by mc2, cistem and relion. Then, as flexalign expects it differently, it's the flexalign protocol which has been modified to make a prior flip of the gain (all this changes apply only when the movies are tiff).
With this new changes, we can ensure an automatic approach for a correct movie alignment gain correction, assuming a correct functioning of movie gain  (same thing that happened with the ones discussed here), but this time there is no need to make changes in other plugins. ( https://github.com/I2PC/scipion-em-xmipp/pull/380 )

This said, any user who decides not to use the movie gain protocol will obviously have to manage the gain orientation by himself, because, as I said, the movie-gain orientations in the databases are few less than random.

About flipping the output micrographs after the flexalign protocol in order to standardize the movie alignment outputs, the code is ready, but I think it is a separated issue and I will send it in another pull request so you can decide about it.

Thank you all for your comments and suggestions.

azazellochg:
This can be closed then?

PR #204, opened 2021-03-29
Tolerate error in movie alignment method when "preprocess fails" 

pconesa:
at least we'll have the mics without plots and protocol does not fail.

Not sure if just having the mic will be an issue later. If so, other option would be to skip the mic.

This is triggered by several K3 acquisitions, that occasionally output movies with 1 less frame.

Optionally, maybe the "preprocessing" method could be modified to tolerate different frame numbers.

### DIFF
--- a/pwem/__init__.py
+++ b/pwem/__init__.py
@@ -39,7 +39,7 @@ from .constants import *
 from .objects import EMObject
 from .utils import *
 
-__version__ = '3.0.8'
+__version__ = '3.0.9'
 _logo = "scipion_icon.gif"
 _references = ["delaRosaTrevin201693"]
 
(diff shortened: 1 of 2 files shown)


azazellochg:
Currently, relion polishing will fail if you have different number of frames.

pconesa:
Should we then skip the mic? or "postpone" the error at polishing. If the latter....is there any way to skip the movie. I guess no.

azazellochg:
Ideally would be to check the header of every movie file for integrity during import like in cryosparc, but I guess this will slow scipion down by a lot. I'd skip such movies. Scipion model doesnt support it. I'll need to check how this is treated by motioncor/polish at the moment.

pconesa:
Motioncor aligns it properly (the main part), but fails when generating the "extra stuff" (plot, alignment). For now I thinks it is safer to just skip the movie.

pconesa:
Done (movie/mic) is skip.

azazellochg (approved):

PR #207, opened 2021-04-09
Dhg cryo assess

DaniDelHoyo:
Allow the function to ask for the output metadata file.
When running the function in parallel, if the output is not specified, the filename is the same and it can trigger concurrence errors

### DIFF
--- a/pwem/protocols/protocol_align_movies.py
+++ b/pwem/protocols/protocol_align_movies.py
@@ -43,6 +43,7 @@ import pyworkflow.protocol.constants as pwcts
 
 import pwem.objects as emobj
 from pwem import emlib
+import xmipp3.utils as xmutils
 
 from pwem.protocols import ProtProcessMovies
 
@@ -232,6 +233,11 @@ class ProtAlignMovies(ProtProcessMovies):
 
             for movie in newDone:
                 mic = micSet.ITEM_TYPE()
+                if hasattr(self, "flipY") and self.flipY:
+                    print('Flipping Y output micrograph')
+                    micPath = self._getExtraPath(getOutputMicName(movie))
+                    newMcFn = xmutils.flipYImage(micPath, outFn=micPath)
+                    movie.setMicName(newMcFn)
                 mic.copyObjId(movie)
                 mic.setMicName(movie.getMicName())
                 # The subclass protocol is responsible of generating the output
--- a/pwem/protocols/protocol_align_movies.py
+++ b/pwem/protocols/protocol_align_movies.py
@@ -525,7 +525,7 @@ class ProtAlignMovies(ProtProcessMovies):
                env=eman2.Plugin.getEnviron())
 
     def averageMovie(self, movie, inputFn, outputMicFn, binFactor=1, roi=None,
-                     dark=None, gain=None, splineOrder=None):
+                     dark=None, gain=None, splineOrder=None, outxmd=None):
         """ Average a movie (using xmipp) taking into account the
          possible shifts and other alignment parameters.
          Params:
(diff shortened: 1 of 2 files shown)


azazellochg:
there will be no xmipp code in cryoassess plugin!

pconesa:
I'm sure you have good reasons for not adding Xmipp code to cryoassess, but
is always good to argue why instead of a
"There will not be Xmipp code".

We all need to back out decisions with reasons otherwise, if all keep the
same approach all this is not going to work.

I've been involved somehow in this and didn't realize the Xmipp
dependency.  Which is harmless since Xmipp is already a dependency.

But what about having an averaging method in the imageHandler?....it could
be diverted to Xmipp or use mrcfile.

azazellochg:
OK, in this case I will not maintain this plugin anymore. I leave it for
Scipion /xmipp team

delarosatrevin:
@pconesa I completely support @azazellochg policy avoiding Xmipp dependency in plugins.  I don't think it is "harmless", since Xmipp development/release cycle/maintenance is a bit complex/caotic.  It is far from being a standard library, easy to install and distribute in Python environments. I  think I we want to try to attract more users, we should aim to avoid dependency on Xmipp (or other packages) and rely on more standard libraries in the core modules.

pconesa:
I agree with you both. I'm just pointing out that we should be ready for
disagreements, bad approach, wrong designs and she. This happens we all
should expose, as you've just done, the reasons why instead of being
categorical and strict with our position.

Let's find a way to get an agreement.

When I say harmless I meant that dependency to Xmipp is there already,
wether we like it or not. Once there is not like other dependencies, that
will require an extra installation. In that sense is harmless to the user
that will not require to install anything else.

I suggested also to have an stack average method in the image handler?

On our side, probably we should have talk / inform Gregory about this
development in advance. I can see it could be shocking to see the PR
without having been informed in advance.

DaniDelHoyo:
Hello everyone, first of all, I am sorry to have caused this issue. 
Of course I planned to inform @azazellochg before doing any pull request on the cryoassess plugin, that's why I did raised this current pull request (in scipion-em and just with general changes), but I did not yet in the cryoassess plugin, which I planned for this Monday. I did pusshed a branch in cryoassess, but without the intention to ask for a merge yet, I am sorry again if that is why this started.
I think we can figure something out which gets everyone happy. For example, I used the xmipp software to average the movies (and correcting for the gain), but there might be other software which does this and causes less trouble. If that is the case I will be happy to implement it, as every other solution you may find out. 

Kind regards to everybody and have a good day.

cossorzano:
Dear all, I do not mind calling Xmipp or not to perform the task. What is your suggestion for calculating an average while correcting with the gain of the frames?

pwem/protocols/protocol_align_movies.py
```diff
@@ -563,6 +563,9 @@ def averageMovie(self, movie, inputFn, outputMicFn, binFactor=1, roi=None,
         if gain is not None:
             args += ' --gain ' + gain
 
+        if outxmd is not None:
```
pconesa, line 566: What happens if outxmd is not passed?

DaniDelHoyo: If the outxmd filename is not passed, the function behaves as before these changes: the outxmd name will be the default (out.xmd if I recall correctly), which can be problematic when running in parallel.

pconesa: That is why y was asking....does is maybe appears at PROJECT_HOME?

I was thinking to have a default behaviour more convenient, like working in the protocol/tmp path and maybe having a timestamp based name to avoid collisions or something based on the move name?

pconesa:
I'm accepting this since is does not harm and allow for to define where xmd will be placed instead of default PROJECT_ROOT/out.xmd

pconesa (approved):

PR #209, opened 2021-04-15
modify function fetch_emdb_map so retries if there is an error

rmarabini:
function "fetch_emdb_map" retrieves a 3D map from PDB server. Sometimes the connection between scipion and EMDB is not great and the function may fail.
I modified it so it tries nTimes (by default nTimes=3) before raising an exception. 

If you want to test it uncomment the line 

            # raise Exception("test")  # uncomment to test this loop

Implementation note:  Loop statements may have an else clause; it is executed when the loop terminates through exhaustion of the iterable

### DIFF
--- a/pwem/protocols/protocol_import/volumes.py
+++ b/pwem/protocols/protocol_import/volumes.py
@@ -421,17 +421,28 @@ def fetch_emdb_map(id, directory, tmpDirectory):
     minimum_map_size = 8192  # bytes
     url_rest_api = url_rest_api % id
 
-    try:
-        map_path, samplingAPI, originAPI = fetch_file(map_url,
-                                                      url_rest_api,
-                                                      name,
-                                                      minimum_map_size,
-                                                      directory,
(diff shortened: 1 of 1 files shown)


fonsecareyna82 (approved):

pconesa (approved):

pconesa:
Maybe, not now, at some point we want to add a dependency "retry" package ( https://pypi.org/project/retry/) so is much easier to make retries (just a decorator).

mmmtnez:
Hi all, 

probably the new modification is affecting the test:

./scipion3 test xmipp3.tests.test_protocols_localdeblur.TestLocalDeblur

since it is failling when imports the volume from EMDB (ID code 3488).
Please review this issue. Thank you

mmmtnez:
Hi all,

after updating Scipion and Xmipp the test mentioned before does not fail in my computer. It only fails in builbot. Could it be anything in buildbot related with the map importing process from EMDB now?

PR #213, opened 2021-04-21
"set filter" protocol and wizard

pconesa:
protocol tree (left pannel) collapsed at loading
protocols.conf created taken from scipion-app defining its own protocols.
version bumped

set filter does not work in streaming now (we might want that)
wizard result is the raw result from the eval which for booleans are 1 or 0 instead of True, False....we might want to make it more intuitive. Also, wizard deals with the first item but we might want to "navigate" through some items to test the formula with more items. A help also for non dev users might be useful too.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,3 +1,9 @@
+V3.0.10
+users:
+  "set filter" protocol and wizard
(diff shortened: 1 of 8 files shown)


cossorzano (approved):
Excellent!! In streaming next :-)

PR #216, opened 2021-05-06
fixes in ctfStr astigmatism definition

fede-pe:
This PR includes:

- Fixes in CTFModes,  __str__ method. Astigmatism is equal to the defocus difference (no defocusAngle)

### DIFF
--- a/pwem/objects/data.py
+++ b/pwem/objects/data.py
@@ -140,7 +140,7 @@ class CTFModel(EMObject):
                  "psh={}, res={}, fit={}".format(
             strEx(self._defocusU.get()),
             strEx(self._defocusV.get()),
-            strEx(self._defocusU.get()-self._defocusV.get()),
+            strEx(self._defocusU.get(0)-self._defocusV.get(0)),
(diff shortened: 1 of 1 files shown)


azazellochg:
Though this is just a str method, in the current data model, astigmatism = DefocusRatio

pconesa:
image
Maybe we should be consistent with other parts. HTML repo shows both, and astigmatism seems to be DU-DV ?

pconesa:
Also, if we do the subtraction, it should be safe, what if one defocus, for some reason is None. We do not want this method to fail because of that.

azazellochg:
I fully support using the difference instead of the ratio and this was raised by @rmarabini 5 years ago :) Just worries about backwards compatibility.

fede-pe:
I have no preference for either option. Actually, I used the difference based on the definition given in the CTFFIND4 paper. Since this is an str method, where could be a problem of backwards compatibility?

azazellochg:
I was referring to the data model in general, this PR is of course harmless.

JorMaister (approved):

PR #223, opened 2021-05-24
Protect defocus ratio calculation in standarize method (CTFModel)

fede-pe:
This PR includes:

- Changes in standardize method in CTFModel to protect from potential errors in defocus ratio calculation (division by cero)

### DIFF
--- a/pwem/objects/data.py
+++ b/pwem/objects/data.py
@@ -122,6 +122,7 @@ class Acquisition(EMObject):
 class CTFModel(EMObject):
     """ Represents a generic CTF model. """
 
+    DEFOCUS_V_MINIMUM_VALUE = 0.1
     DEFOCUS_RATIO_ERROR_VALUE = -1
 
     def __init__(self, **kwargs):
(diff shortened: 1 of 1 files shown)


pwem/objects/data.py
```diff
@@ -251,7 +253,11 @@ def standardize(self):
             self._defocusAngle.sum(180.)
         # At this point defocusU is always greater than defocusV
         # following the EMX standard
-        self._defocusRatio.set(self.getDefocusU() / self.getDefocusV())
+
+        if self.getDefocusV() != 0:
```
delarosatrevin, line 257: It is not a good idea to use equality (or inequality) comparison on floating-point values. Maybe it should be tested with a given very small error (i.e close to zero)

fede-pe: Done. I set the minimum of DefocusV as 0.1, do you think it is restrictive enough?

pconesa (approved):

fonsecareyna82 (approved):

PR #224, opened 2021-05-26
New 3D Wizards for Mask/Origin Selection

DavidHerreros:
Two new wizards have been added to the `wizards/wizards_3d` folder including the following functionalities:

- `mask_strcuture_wizard`: This wizard opens an `Atomic Structure` with Matplotlib to define a spherical mask or an origin. The values returned by the wizard are the radius of the sphere defining the mask and the origin of the sphere (it can return both or just one depending on the application).

-  `mask_volume_wizard`: This wizard opens a `Volume` with Matplotlib to define a spherical mask or an origin. The values returned by the wizard are the radius of the sphere defining the mask and the origin of the sphere (it can return both or just one depending on the application).

By default, the wizard returns the origin of the sphere following the Xmipp convention (coordinates x,y,z are referred to the center of the `Atomic Strcuture`/`Volume`).  In the `Volumes` implementation, it is also possible to return the origin following the Bottom Left Convention from Eman.

If working with `Volumes`, the units of the radius/origin are in voxels.

### DIFF
--- a/pwem/wizards/wizards_3d/__init__.py
+++ b/pwem/wizards/wizards_3d/__init__.py
@@ -1,6 +1,6 @@
 # **************************************************************************
 # *
-# * Authors: Yunior C. Fonseca Reyna    (cfonseca@cnb.csic.es)
+# * Authors: David Herreros Calero    (dherreros@cnb.csic.es)
 # *
 # *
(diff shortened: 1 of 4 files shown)


EstrellaFG (approved):
Good job and very nice colors! ;D

pconesa (approved):

PR #227, opened 2021-06-08
Fixed atomic structure output names in test

mmmtnez:
Hi all,
here you have some small changes to fix the atomic structure output names according to the protocol number since it started to fail after some scipion3 update. Only the test has been modified and you can check that it works properly with:

scipion3 tests pwem.tests.workflows.test_workflow_modeling

### DIFF
--- a/pwem/tests/workflows/test_workflow_modeling.py
+++ b/pwem/tests/workflows/test_workflow_modeling.py
@@ -1007,10 +1007,13 @@ class TestCootRefinement(TestImportData):
             self.launchProtocol(protCoot)
         except Exception as ex:
             print("second call to coot ended")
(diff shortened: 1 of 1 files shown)


rmarabini:
Hi @mmmtnez,
     
     I think I should not decide on this pull request.  So @pconesa,  it is up to you. Let me comment the problem that this pull request addresses.

   Some of our output filenames contain the number (id) of the protocol. On a test this numbers are  always the same no mater how many times the test is run unless... scipion core is updated. In that case, if some extra objects are created, the protocol ids change.

  In summary, in the test we can not assume that the protocol id is going to be always the same but we need to compute it. This pull request introduces that modification.

pconesa (approved):
I see not a problem with this. Much robust this way.

PR #235, opened 2021-07-13
Rm seteditor

rmarabini:
Modification of the protocol "metadata editor". This protocol allows to modify the metadata files.

Introduced modifications:

1) rename from metadata-editor to set-editor. Since there is another  protocol called set-filter, this new name seems to be more consistent. Furthermore, the protocol can only modify setsOfThings

2) Add test 
   scipion3 tests pwem.tests.protocols.test_protocols_set_editor

3) Add the possibility to apply a rotation to the alignment matrix. Follows a description of the problem that I tried to solve here. A given symmetry may be defined in may ways (for example icosahedral symmetry can be defined as i1, i2, i3,...) sometimes a volume (and the corresponding projection matrices) reconstructed using symmetry i1 needs to transformed to other symmetries (i.e i2). This extension solves that problem.

Note: point (3) could be carried on using a very long formula with the previous version of the protocol, but I think I am the only user able to write the formula.


pwem/protocols/protocol_set_editor.py
```diff
@@ -0,0 +1,282 @@
...
+        modifiedSet = inputSet.createCopy(self._getExtraPath(), copyInfo=True)
+        for sourceItem in inputSet.iterItems():
+            item = sourceItem.clone()
+            transformation = item.getTransform()
+            transformation.composeTransform(matrix)
+            modifiedSet.append(item)
+        self.createOutput(self.inputSet, modifiedSet)
+
+    def createOutput(self, inputSet, modifiedSet):
+        # TODO: getExtendend is empty when protocol called
```
pconesa, line 224: There are 2 ways to define a pointer, a direct pointer, and an indirect one.

Through the gui is not possible to have a direct pointer (point directly to an id that is the output itself)
Only from tests we create direct pointers

One way is to create indrect pointers in the test like:

prot.inputSet.set(protWithOutput)
prot.inputSet.setExtended("outputSet")

pconesa: You want to keep tha same name as the input?

pwem/tests/protocols/test_protocols_set_editor.py
```diff
@@ -0,0 +1,220 @@
...
+                                   doCtf=True)
+        protImportProj   = self.importData(setPartSqliteName,
+                                           "import projection\n operation",
+                                           ProtImportParticles,
+                                           ProtImportParticles.IMPORT_FROM_SCIPION)
+
+        #launch operate set protocol
+        protSetEditor = self.newProtocol(ProtSetEditor,
+                                         objLabel="operate")
+        protSetEditor.inputSet.set(protImportProj.outputParticles)
```
pconesa, line 156: Here, you can do 
protSetEditor.inputSet.set(protImportProj)
protSetEditor.inputSet.setExtended("outputParticles")

rmarabini:
OK, I will fix this direct/inderect pointer

pconesa (approved):

rmarabini:
I modified the files following @pconesa  advice about how to use "extendend"

PR #238, opened 2021-07-27
Try to get CHIMERA_HOME from chimera plugin.

pconesa:
This PR tries to get the CHIMERA_HOME from the chimera plugin itself. This way any future chimeraX update will not require to release this code.

In the future, maybe code here using chimera could be moved to chimera plugin, but this is not done here.

Increase JAVA max memory to 4GB


rmarabini (approved):
OK with me. I will merge also scipion-em-chimera

PR #239, opened 2021-07-30
link .gain as .tif file before conversion

azazellochg:
.gain gain reference file from Falcon 4 / EPU is a tif file. Since xmipp cant recognize it as tif (but rather unknown spider file), I make a tmp link before actual conversion to mrc. The link has to be absolute for cases when tmp is on separate scratch disc


azazellochg:
Can anyone review this?

pconesa:
upss! I missed this. Sorry.

pconesa (approved):

PR #248, opened 2021-08-19
Fix Z rot matrix

JorMaister:


pconesa (approved):

azazellochg:
Hi,

this lokos like a small but very serious change. Has this been verified / tested extensively? How does this affect packages?

azazellochg:
Ah, I see this create func is only used in jjsoft plugin. Alright then.

pconesa:
Hi, this code is was recently added and I doubt any existing code is using it, except for the one, in tomography that might be using it and triggered the creation of this methods.

PR #253, opened 2021-09-24
Number and proportion of filtered items

DaniDelHoyo:
Allow the filtering of a specified number/proportion of items with the highest/lowest values of a given attribute

### DIFF
--- a/pwem/protocols/protocol_set_filter.py
+++ b/pwem/protocols/protocol_set_filter.py
@@ -242,7 +242,7 @@ class ProtSetFilter(EMProtocol):
         elif self.topRankOrder.get() == self.LOWER:
             direc = 'ASC'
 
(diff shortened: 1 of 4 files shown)


pwem/protocols/protocol_set_filter.py
```diff
@@ -199,11 +225,52 @@ def formulaStep(self):
...
+            if i in itemsIdx:
+                modifiedSet.append(item.clone())
+                added += 1
+            if added == finalNumber:
+                break
+
+        self.createOutput(modifiedSet)
+
+
+    def getAttributeValues(self, attribute, iSet):
```
pconesa, line 262: this implies reading all set and putting it in a list. Should be done differently. Using the arguments in iterItems: orderby and maybe limit

pconesa (approved):

PR #257, opened 2021-09-27
Added an small molecules dataset

DaniDelHoyo:


JorMaister (approved):

pwem/tests/__init__.py
```diff
@@ -283,3 +283,11 @@ def defineDatasets():
                 'allMics': 'micrographs/*.mrc',
                 'autoPickingCoordinates': 'autoPickingCoordinates'
             })
+
```
pconesa, line 286: Datasets can be define inside plugins. This is not the right place for this one. ;-)

DaniDelHoyo: Hi Pablo, I tried defining it inside the scipion-chem plugin but seems like it didn't work.
I guess the datasets definition goes over some plugins but maybe not over the scipion-chem, that's why I added it here.
Should we change it otherway?

pconesa:
Have you tried this:

https://github.com/scipion-em/scipion-em-continuousflex/blob/master/continuousflex/__init__.py#L129

DaniDelHoyo:
Okay, defining it in the plugin/__init__() instead of in the plugin/tests/__init__() did work. Thank you

PR #261, opened 2021-10-15
sqlite with ..

DaniDelHoyo:
The output sqlite database generated with the previous code was "pockets..sqlite" (note the double point). 
should it be fixed in the createCopy method?


DaniDelHoyo:
Today I was somehow braver, it itched me every time I saw this open PR.

PR #266, opened 2021-11-10
Fix outputClassName in import protocols

pconesa:


pconesa:
Fixes an error introduced after using classes in possible outputs instead of instances

danielmarchan3 (approved):

fonsecareyna82 (approved):

PR #270, opened 2021-11-19
Possible output os movie alignment changed to dictionary.

pconesa:
Turns out that Enum can't have two items with the same value and keep them separately.


azazellochg (approved):
Thanks a lot, @pconesa, for finding and fixing it!

fonsecareyna82 (approved):

PR #274, opened 2021-12-15
Methods for writing fasta files for Sequence and SetOfSequences

DaniDelHoyo:
Added methods for writing fasta files for Sequence and SetOfSequences

### DIFF
--- a/pwem/convert/sequence.py
+++ b/pwem/convert/sequence.py
@@ -53,7 +53,7 @@ import sys
 from Bio.SeqRecord import SeqRecord
 from Bio.Align.Applications import ClustalOmegaCommandline, MuscleCommandline
 from Bio import pairwise2
-
+from pyworkflow.utils import getExt
 
 class SequenceHandler:
     def __init__(self, sequence=None,
@@ -70,19 +70,70 @@ class SequenceHandler:
             self._sequence = None
             # type(self._sequence):  <class 'Bio.Seq.Seq'>
 
-    def saveFile(self, fileName, seqID, sequence=None, name=None,
+    def getTypeFromFile(self, fileName):
+        '''Returns the expected BioPython file type according to the filename'''
+        ext = getExt(fileName)
+        if ext == '.fasta':
+            type = 'fasta'
+        elif ext == '.genbank' or ext == '.gb':
+            type = 'genbank'
+        else:
+            type = 'fasta'
+        return type
+
+    def appendFile(self, fileName, seqID, sequence=None, name=None,
                  seqDescription=None, type="fasta"):
+        '''Appends a sequence to the sequences in a file'''
         if sequence is not None:
             self._sequence = sequence
-        record = SeqRecord(self._sequence, id=seqID, name=name,
-                           description=seqDescription)
+
+        if type is None:
+            type = self.getTypeFromFile(fileName)
+
+        if os.path.exists(fileName):
+            records = list(SeqIO.parse(fileName, type))
+        else:
+            records = []
+        records.append(SeqRecord(self._sequence, id=seqID, name=name,
+                                 description=seqDescription))
+        with open(fileName, "w") as output_handle:
+            SeqIO.write(records, output_handle, type)
+
+    def saveFile(self, fileName, seqID, sequence=None, name=None,
+                 seqDescription=None, type="fasta"):
+        '''Saves a single sequence into a specified file'''
+        if sequence is not None:
+            self._sequence = sequence
+
+        self.saveFiles(fileName, [seqID], [self._sequence], [name], [seqDescription], type)
+
+    def saveFiles(self, fileName, seqIDs, sequences, names,
+                  seqDescriptions, type="fasta"):
+        '''Saves multiple sequences into a specified file'''
+        if type is None:
+            type = self.getTypeFromFile(fileName)
+
+        records = []
+        for seq, seqID, name, seqDescription in zip(sequences, seqIDs, names, seqDescriptions):
+            records.append(SeqRecord(seq, id=seqID, name=name,
+                               description=seqDescription))
         # type(record): < class 'Bio.SeqRecord.SeqRecord'>
         with open(fileName, "w") as output_handle:
-            SeqIO.write(record, output_handle, type)
+            SeqIO.write(records, output_handle, type)
 
     def downloadSeqFromFile(self, fileName, type="fasta"):
         record = next(SeqIO.parse(fileName, type))
         return record
+
+    def readSequencesFromFile(self, fileName, type='fasta'):
+        '''From a sequences file, returns a list of dictionaries with each sequence info.
+        Dictionary: [{'seqID': seqID1, 'sequence': sequence1, 'description': description1},
+                     ...]'''
+        sequences = []
+        records = SeqIO.parse(fileName, type)
+        for rec in records:
+            sequences.append({'seqID': rec.id, 'sequence': rec.seq, 'description': rec.description})
+        return sequences
 
     def downloadSeqFromDatabase(self, seqID):
         # see http://biopython.org/DIST/docs/api/Bio.SeqIO-module.html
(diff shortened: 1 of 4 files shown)


edat-uam-es:
I am not very happy with this pull request. With pdb and sequence data we have always try to follow the policy that we do not do a direct access to data but let well established libraries to handle it. 

Sequence IO is centralized in "SequenceHandler" which is able to read/write many formats including fasta. Once the sequence is read the original format is lost and became a python object which, if needed, may be saved again in fasta format using "SequenceHandler" 

So, in summary:

1) IO should not be in data.by
2) IO should be in sequencehandler
3) Once the sequence is read it became a python object which does not have fasta header but that could be saved in fasta format.

rmarabini:
I am not very happy with this pull request. With pdb and sequence data we have always try to follow the policy that we do not do a direct access to data but let well established libraries to handle it. 

Sequence IO is centralized in "SequenceHandler" which is able to read/write many formats including fasta. Once the sequence is read the original format is lost and became a python object which, if needed, may be saved again in fasta format using "SequenceHandler" 

So, in summary:

1) IO should not be in data.by
2) IO should be in sequencehandler
3) Once the sequence is read it became a python object which does not have fasta header but that could be saved in fasta format.

rmarabini (changes requested):
I  am not very happy with this pull request. With pdb and sequence data we have always try to follow the policy that we do not do a direct access to data but let well established libraries to handle it. 

Sequence IO is centralized in "SequenceHandler" which is able to read/write many formats including fasta. Once the sequence is read the original format is lost and became a python object which, if needed, may be saved again in fasta format using "SequenceHandler" 

So, in summary:

1) IO should not be in data.by
2) IO should be in sequencehandler
3) Once the sequence is read it became a python object which does not have fasta header but that could be saved in fasta format.

DaniDelHoyo:
Okay, I understand your point, that was why I asked Marta first so I didn't mess with you objects, sorry I didn't do so with you too. So, some questions:

Do you think it would be ok to have the writeFasta method in the Sequence object but that this method uses the SequenceHandler instead of using the "dummy code" I did?

For the SetOfSequences, is there an option in SequenceHanlder / biopython to write combinedFasta files (one file, several sequences) ?

About the new method I added to the PR, guessIsAminoacid, maybe it is also a dummy method, is there a problem if the alphabet is not the standard? (I don't know much about the different alphabets used for sequences)

These methods would be quite useful for me but as I said, I don't want to mess with your code. If there is no way to agree on this, I can always make my own objects, sons of yours, so your code keeps untouched.

rmarabini:
Hi
> Okay, I understand your point, that was why I asked Marta first so I didn't mess with you objects, sorry I didn't do so with you too. So, some questions:
> 
> Do you think it would be ok to have the writeFasta method in the Sequence object but that this method uses the SequenceHandler instead of using the "dummy code" I did?
> 
I would suggest to create in the file data.py a method named  exportToFile().

The code of such function would be something similar to (careful, I have not check the code):

def exportToFile(self, seqFileName): # self is a sequence object
        import pwem.convert as emconv
        seqHandler = emconv.SequenceHandler(self.getSequence(),
                                            isAminoacid=self.getIsAminoacids())
        seqBio = seqHandler._sequence  # Bio.Seq.Seq object
        
        # retrieving  args from scipion object
        seqID = self.getId()
        seqName = self.getSeqName()
        seqDescription = self.getDescription()
        seqHandler.saveFile(seqFileName, seqID, sequence=seqBio,
                            name=seqName, seqDescription=seqDescription,
                            type=None)

and modify seqHandler.saveFile(...,type='fasta') so if type==None then the method tries to guess the filetype by the extension.

        ext = getExt(seqFileName)
        if type is None:
            if ext == '.fasta':
                  type = 'falta'
           elif ext = ....

           else:
                type='fasta'

See https://biopython.org/wiki/SeqIO for al the sequence types defined in biopython

> For the SetOfSequences, is there an option in SequenceHanlder / biopython to write combinedFasta files (one file, several sequences) ?
> 

As it is implemented right now SequenceHanlder does not support this option but biopython does..  
Nevertheless, be aware that a sequence should NOT be able to save a setOfSequences. So the first step would be create a new scipion class called setOfSequences and modify SequenceHandler so "self.sequence" became a list. Of course, we assume that all sequences have the same alphabet and kind (Aminoacid/not aminoacid)


Note that in  seqHandler.saveFile() you may find the line

      SeqIO.write(record, output_handle, type)

where record is a sequence. If you pass here a List of sequences all of then will be saved in the same file.


> About the new method I added to the PR, guessIsAminoacid, maybe it is also a dummy method, is there a problem if the alphabet is not the standard? (I don't know much about the different alphabets used for sequences)
> 
Uhmmm, I do not think guessIsAminoacids should be in data.py. this information should be available in the sequence object. Whatever you use to import the sequence should solve de problem.

> These methods would be quite useful for me but as I said, I don't want to mess with your code. If there is no way to agree on this, I can always make my own objects, sons of yours, so your code keeps untouched.

Most definitely, it should not be two classes with the same functionality that will be the end of Scipion as an integration package. The classes are not  yours or mine,  there are part of scipion and in case of disagreement we should ask for advice from others to integrate them properly. Note that in sequence we are trying to follow the very same approach used for images and atomic models.
If the data model for sequence diverges then the interoperability between protocols is not longer guarantied.

  If by any chance you are at CNB we may discuss this topic face to face. I am sure that will help to clarify the different points of view.

DaniDelHoyo:
Si! Estoy en el B13. Si prefieres puedo ir donde estés

DaniDelHoyo:
I have added in appendFile a check for if the file to append exists (if not, it just creates the file and saves the sequence on it).
SetOFSequences have its own export to file too which uses this appendFile

rmarabini:
Almost there ;-). I see that there is a little inconsistency between the methods

readSequencesFromFile, downloadSeqFromFile and downloadSeqFromDatabase

The first one is new while the second and third ones where created either by Marta o by myself.

I like the name readSequencesFromFile but then downloadSeqFromFile should be renamed to readSequenceFromFile (singular) since both methods do almost the same. Also downloadSeqFromFile should not implement the reading but call to  readSequencesFromFile and return the first element of the list.

readSequencesFromFile returns a list of dictionaries  while downloadSeqFromFile and downloadSeqFromDatabase return a record (Biopython structure). I think the new proposed structure (dictionary) is a better option that the old one (biophython record) since  encapsulates biopython in the file convert/sequence.py. Would it be possible to modify  downloadSeqFromFile and downloadSeqFromDatabase so they return a dictionary.   I did a bit of research and  these two functions are only used in import_sequence and in convert/sequence.py so the corresponding code can be easily updated. I attach the lines that need to be modified

scipion3/scipion-em/pwem/protocols/protocol_import/sequence.py:        record = seqHandler.downloadSeqFromFile(sequenceFile,
scipion3/scipion-em/pwem/convert/sequence.py:    return len(handler.downloadSeqFromFile(filename, format))
scipion3/tmp/scipion-em/pwem/protocols/protocol_import/sequence.py:        record, error = seqHandler.downloadSeqFromDatabase(sequenceDB)


With these two extra changes I will completely and absolutely happy ;-)

DaniDelHoyo:
That should be it then, I have added 'alphabet' to the dictionary and converted 'sequence' to string, to make sure no BioPython object is passed

DaniDelHoyo:
Just one more thing, would it be okay to make a method for sequence which calls the readSequenceFromFile so we don't have to instantiate a SequenceHandler to use it? Just as exportToFile, this time importFromFile

DaniDelHoyo:
It would be something like this, I tried to follow the same structure as for exporting

DaniDelHoyo:
I think it is finally finished, I'll stay tuned for your answer

rmarabini (approved):

rmarabini:
Hi,

  This is OK with me. I have approved the changes but since scipion-em has been modified I think @pconesa  should be the one pushing the "Merge Pull Request" button.

pconesa:
I think this is perfectly fine. Nevertheless, since we are about to release code, I'd rather leave it in devel for the next release to mature, like the good wine!

DaniDelHoyo:
I'm perfectly fine with that, just to know it, when would the release be?

pconesa:
Today! we hope

PR #280, opened 2022-01-03
viewer SetOfAtomStructs

DaniDelHoyo:
Viewer for SetOfAtomStruct object.

### DIFF
--- a/pwem/viewers/viewer_chimera.py
+++ b/pwem/viewers/viewer_chimera.py
@@ -36,6 +36,7 @@ import pwem.emlib.metadata as md
 import pwem.objects as emobj
 from pwem import emlib, Domain
 from pwem import Config as emConfig
+from pwem.objects import SetOfAtomStructs
 
 chimeraPdbTemplateFileName = "Atom_struct__%06d.cif"
 chimeraMapTemplateFileName = "Map__%06d.mrc"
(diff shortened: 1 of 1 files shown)


rmarabini (changes requested):
Hi,

   I think the code is correct but there is a lot of repetition. I mean the code for  atomStruct and for a setOfAtomStructs is  almost identical.
  Can you:
1) check if the input object is an atomicStuct
2) If affirmative create a temporary setOfAtomicStructs with a single  atomStruct
3) Delete all the code related with viewing a single AtomStruct
4) Keep the code for viewing a setOfAtomStructs

DaniDelHoyo:
There were a couple of differences from the viewer of AtomStruct and the one I took the code from (_displayMapModel from PhenixProtRefinementBaseViewer), I marked them so we can decide what to do

rmarabini:
hi, 

  Let see..

(1)
The targets of the viewer are:   

    _targets = [emobj.AtomStruct, emobj.PdbFile, emobj.SetOfAtomStructs]

but I do not see the code that processes  emobj.PdbFile. emobj.PdbFile is the old name of the class emobj.AtomStruct so may be change line 327 to 

            if issubclass(cls, emobj.AtomStruct) or issubclass(cls, emobj.PdbFile):

will be enough. emobj.PdbFile inherits from emobj.AtomStruct

(2) 

           # FIXME: there is an asymmetry between ProtocolViewer and Viewer
            # for the first, the visualize method return a list of View's
            # (that are shown)
            # for the second, the visualize directly shows the objects.
            # the first approach is better

It is OK with me if we always return a list of views

 (3) I do not like to use the /tmp directory

                 objSet = SetOfAtomStructs.create('/tmp')

it will be nicer to use the protocol tmp directory but... it is deleted after the protocol finished so I guess there is no way around here :-(. I do not know how setOfAtomStruct work, what are the names of the files stored in "/tmp". If two users are using the  viewer, can one user overwrite the file created by the other?

DaniDelHoyo:
(1), as pdbFile is a subclass of AtomStruct, the issubclass function should detect a pdbFile and get into the condition so the code as it is should work.

(2) the fixme was there before I arrived, do you think we should delete it?

(3) True, I have added the protocol Pid as suffix so if several viewers are opened simultaneusly, there should be no problem.

pwem/viewers/viewer_chimera.py
```diff
@@ -316,61 +317,63 @@ def show(self):
...
+                    #f.write("style stick\n")
+
                 f.close()
-                ChimeraView(fnCmd).show()
-            # FIXME: there is an asymmetry between ProtocolViewer and Viewer
-            # for the first, the visualize method return a list of View's
-            # (that are shown)
-            # for the second, the visualize directly shows the objects.
-            # the first approach is better
+                # run in the background
```
pconesa, line 372: Visualize method should return a view not running anything. Although it may work is not the right way to do it. 

Not important, but in the past, when the webtools were available, and in case in the "distant" future we provide a web interface, this runChimeraProgram is a problem.

pconesa (changes requested):

rmarabini:
> (1), as pdbFile is a subclass of AtomStruct, the issubclass function should detect a pdbFile and get into the condition so the code as it is should work.

You are right

> 
> (2) the fixme was there before I arrived, do you think we should delete it?
> 
see @pconesa's comment

> (3) True, I have added the protocol Pid as suffix so if several viewers are opened simultaneusly, there should be no problem.

OK

DaniDelHoyo:
Update:
 (2): the visualize method now returns a view (I therefore deleted the FIXME comment, please confirm this is right)
 (3): I used the ObjId as suffix instead of this pId which is less intuitive (at least to me, I use to work with the objid)

rmarabini:
Hi, this is OK with me

rmarabini (approved):

pconesa (approved):

PR #282, opened 2022-01-11
Rmsd atom

DaniDelHoyo:
RMSD for SetOfAtomStruct protocol and viewer


DaniDelHoyo:
I placed the viewer together with the localRes viewer as it resembles its function (but with AtomStructs instead of volumes), but I can just move it somewhere else if preferred.

DaniDelHoyo:
Here I have some screenshots of what the viewer looks lik
Screenshot from 2022-01-14 13-36-44
Screenshot from 2022-01-14 13-39-19
e

DaniDelHoyo:
I understand, I also think that would be a better option. I can learn how to work with mmCIF files and create them in these protocols in order to store the attributes in them. Also, do you know if chimera has any way to define attributes directly from this mmCIF columns?

The only problem I can see is that an attribute would be lost if we use this mmCIF in a protocol which converts it to PDB, as the attributes are only stored there. I don't know if this situation may happen too much tho.

DaniDelHoyo:
With this last commit, the attributes are stored in the CIF file and the viewer works with this new configuration.
I have stored the attributes in a new section of the CIF tho, I have just seen the mail where you suggest we should do it in the _atom_site section, but sometimes they are not related to atoms but to residues (or even residue type), so it would need some coding to "translate" it to atoms.

rmarabini:
Could you (1) se  chimeraX "key" command to create the colorkey
                 (2) add a test fr each protocol / function added
                 (3) move protocol rmsd_protocol outside scipion-em. May be scipion-em-atomstruct.

DaniDelHoyo:
punto (2) hecho, ya hay un test para la funcion addScipionAttribute de convert.
También he añadido un protocolo para importar sets de atomstruct, con su correspondiente test

rmarabini (approved):
This seem to be OK but @pconesa is the person to push the merge buttom.
By the way, I have been lazy I  did not download the code and run the tests. :-(

pwem/viewers/viewer_localres.py
```diff
@@ -108,3 +117,213 @@ def createChimeraScript(self, scriptFile, fnResVol,
...
+        f.write("run(session, 'volume #%d style surface voxelSize %f')\n" % (1, sampling))
+        f.write("run(session, 'volume #%d origin %0.2f,%0.2f,%0.2f')\n" % (1, x, y, z))
+        f.write("run(session, 'hide #!%d models')\n" % 1)
+
+      else:
+        dim, sampling = 150., 1.
+        strId = 2
+
+      tmpFileName = os.path.abspath(prot._getExtraPath("axis_input.bild"))
+      Chimera.createCoordinateAxisFile(dim,
```
pconesa, line 222: What if Chimera plugin is missing?

pwem/viewers/viewer_localres.py
```diff
@@ -108,3 +117,213 @@ def createChimeraScript(self, scriptFile, fnResVol,
...
+            attrValues = list(map(float, attrValues))
+            return min(attrValues), max(attrValues)
+
+    def getColors(self):
+      low, high = self.getValuesRange()
+      stepColors = splitRange(high, low, splitNum=self.intervals.get())
+      colorList = plotter.getHexColorList(len(stepColors), self.colorMap.get())
+      return stepColors, colorList
+
+    def generateColorLegend(self, stepColors, colorList):
```
pconesa, line 296: Could we use this approach in the resolution viewers?

pconesa: Wait! this is the place... I'm not seeing it when executing.

pwem/viewers/viewer_localres.py
DaniDelHoyo: this problem was present already as the LocalResolutionViewer also uses some Chimera funstions.
Should we make more robust the import for both of them (and maybe launch a warning window for both of the viewers if chimera is missing?)

pconesa (approved):

PR #310, opened 2022-03-19
add even/odd support for SPA motioncor

azazellochg:
Currently used only by mc2


pconesa (approved):
I think is a neat change, with possible outputs updated and harmless. Good job!

fonsecareyna82 (approved):

PR #316, opened 2022-04-11
New mathematical operator protocol

danielmarchan3:
Protocol and test to do mathematical operations over numeric set atttributes or a numeric formula.


cossorzano (approved):

pwem/tests/protocols/test_protocol_mathematical_operator.py
```diff
@@ -0,0 +1,110 @@
...
+# ***************************************************************************/
+
+
+
+from pyworkflow.tests import BaseTest, setupTestProject, DataSet
+from pyworkflow.plugin import Domain
+from pwem.protocols.protocol_import import ProtImportMicrographs
+from pwem.protocols.protocol_mathematical_operator import ProtMathematicalOperator
+
+SphireProtCRYOLOPicking = Domain.importFromPlugin('sphire.protocols', 'SphireProtCRYOLOPicking', doRaise=True)
```
pconesa, line 31: We should avoid other plugins dependencies if possible.

In this case, it is easy to make the test using other protocol from this plugin instead of the one from sphire. I suggest to adapt this.

danielmarchan3:
> Nice job Daniel, my only suggestion would be to use an import protocol in the test instead of the sphire one.

I undestand it the thing is that I am using the estimated box size as an inputPointer, but yes I can use instead the result of the manual mathematical operator as new input.

pconesa:
If you want to test the case for having a IntegerOutput you can always fake it in the test,like:

protImportMic = self.launchProtocol(...)
protImportMic.defineOutputs(outputInteger = Integer(3))
protImportMic.store() # or store keep forgetting this.

pconesa (approved):

PR #320, opened 2022-04-14
creating a subset from specific ids

MohamadHarastani:
This will allow creating a subset from scipion objects by setting a range of IDs. I did not modify the form in order not to affect other protocols that depend on this protocol.


pconesa:
Hi @MohamadHarastani , I was toady testing this.

Do you think that "Make random subset" should be hidden when activating the new option?

jamesmkrieger:
> Do you think that "Make random subset" should be hidden when activating the new option?

Yes

jamesmkrieger (approved):
It could be better with Pablo's suggestion of adding a condition on the random subset option, but what you have looks good I think.

PR #321, opened 2022-04-18
Simplification of the mathematical operator protocol.

pconesa:
Just 2 inputs with a formulae.
Future improvements ideas:
- Multiple inputs (but then this will imply multipointer params to Scalars). Is this implemented?
- Instead of eval, use a mor complex multiline code and use compute?

Test adapted.


pconesa:
@danielmarchan3 , this is a proposal to simplify the math operator protocol.

danielmarchan3 (approved):

PR #325, opened 2022-05-09
New method to extract b-factor values

mmmtnez:
Dear all,

I have included in the convert file a new method to extract the B-factor values (or any other type of values of this column, for example LDDT or RMSD) from an atomic structure.


rmarabini (approved):
No problems here. 
Should it be a setStructureBFactorValues method?
                       ^^^

PR #329, opened 2022-05-12
Completed the method getStructureBFactorValues

mmmtnez:
Dear all,

here you have a second more complete version of the method getStructureBFactorValues to get two different lists including B-factor values. In addition to the previous one (list of each atom B-factors), we have a second list that associates each residue with the average of the B-factor values of all atoms in the residue. This second list is in fact a list of lists including for each residue: [Model, chain, residue order number, residue name, average B-factor].


rmarabini (approved):
OK with me

PR #330, opened 2022-05-13
add number to SetOfAS localPath to fix overwriting

jamesmkrieger:
If there are different folders with atomic structure files with the same name that get selected with wild cards then they would overwrite each other previously. 

Now, if there's already something with that name then we add a number e.g. atoms.pdb becomes atoms2.pdb if atoms.pdb is already there.

### DIFF
--- a/pwem/protocols/protocol_import/volumes.py
+++ b/pwem/protocols/protocol_import/volumes.py
@@ -473,24 +473,24 @@ class ProtImportSetOfAtomStructs(ProtImportFiles):
       padding_format_str = '%0' + str(len(str(numStructs))) + 'd'
 
       for atomStructPath in atomStructPaths:
-        if not exists(atomStructPath):
-          raise Exception("Atomic structure not found at *%s*" % atomStructPath)
+          if not exists(atomStructPath):
+              raise Exception("Atomic structure not found at *%s*" % atomStructPath)
 
-        baseName = basename(atomStructPath)
-        localPath = abspath(self._getExtraPath(baseName))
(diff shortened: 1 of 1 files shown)


jamesmkrieger:
There was some problem with indentation because this file uses 2 tabs, but now everything works again

pwem/protocols/protocol_import/volumes.py
```diff
@@ -468,19 +468,29 @@ def createOutputStep(self, atomStructPaths):
...
       """
       outASs = emobj.SetOfAtomStructs().create(self._getPath())
+      
+      numStructs = len(atomStructPaths)
+      padding_format_str = '%0' + str(len(str(numStructs))) + 'd'
+
       for atomStructPath in atomStructPaths:
-          if not exists(atomStructPath):
-              raise Exception("Atomic structure not found at *%s*" % atomStructPath)
+        if not exists(atomStructPath):
```
pconesa, line 476: Indentation does not seem correct?

jamesmkrieger: I was having a lot of problem with the indentation in this file. I'll have another look. Thanks

jamesmkrieger: It should be right now

jamesmkrieger:
I can confirm it works again now too

pconesa:
Thanks @jamesmkrieger test are passing now.

pconesa (approved):

jamesmkrieger:
great. You're welcome

PR #339, opened 2022-07-01
Improve functionality of SetOfSequence.

DaniDelHoyo:
ImportFiles function allows file type as parameter
Set of sequence attribute defines if sequences are aligned (optional)

ChimeraAttributeViewer n (resolution): it wouldn't let you draw an histogram to low  ranges (below ~0.2 interval)


pwem/wizards/wizards.py
```diff
@@ -351,7 +351,7 @@ def show(self, form, *params):
       for residue in finalResiduesList:
           if json.loads(residue.get())['index'] == idxs[0]:
               inSeq = True
-          elif json.loads(residue.get())['index'] == idxs[-1]:
+          if json.loads(residue.get())['index'] == idxs[-1]:
```
pconesa, line 354: why is not elif anymore?

pconesa (changes requested):

PR #340, opened 2022-07-03
Fix protname

azazellochg:


setup.py
```diff
@@ -62,16 +62,18 @@
         #   3 - Alpha
         #   4 - Beta
         #   5 - Production/Stable
-        'Development Status :: 2 - Pre-Alpha',
+        'Development Status :: 4 - Beta',
```
pconesa, line 65: Shouldn't we go for 5?

azazellochg: I was thinking since this is devel branch it should be beta, but probably it's too much hassle to keep a different status for master/pypi. Yep, let's go for 5!

pconesa (approved):

PR #345, opened 2022-07-28
Cuda version guess

pconesa:


azazellochg:
woah, so many functions, looks like overengineering...

pconesa:
Well, half of them could be used for other stuff, are quite generic.

PR #355, opened 2022-09-05
using logging in test data

pconesa:


pwem/tests/data/test_data.py
```diff
@@ -58,7 +58,7 @@ def createDummyProtocol(projName):
     try:
         proj.launchProtocol(prot)
     except Exception as e:
-        print(str(e))
+        logger.error("Can't launch EMProtocol")
```
fonsecareyna82, line 61: Wouldn't it be convenient to print the error too?

pconesa: Done

fonsecareyna82 (approved):

PR #362, opened 2022-09-19
Attributes viewer include viewer of the attribute over the sequence

DaniDelHoyo:
The attributes can be visualized over the sequence, not only over the structure


pconesa (approved):

pconesa:
PLease add a CHANGES.txt line.

PR #367, opened 2022-10-05
Fix alignment assign

pconesa:


pwem/protocols/protocol_alignment_assign.py
```diff
@@ -104,11 +101,21 @@ def _updateItem(self, item, row):
...
     def createOutputStep(self):
+
         inputParticles = self.inputParticles.get()
-        inputAlignment = self.inputAlignment.get()
+
+        # Store data to be used in the update item
+        self.alignmentData = self.inputAlignment.get()
+
+        # Trigger the connection to avoid connecting and disconnecting later when using self.alignmentData[23]
+        self.alignmentData.load()
```
azazellochg, line 111: that's very useful to know!

pconesa: One day I'll make a "performance tips" page. In summary:

Be careful with enumerate(set) or list(set): It will load all set in memory. If you just want the ids or a specific field, use the aggregation functions.

If you use myset[number] .... this may leave the connection in the same status as it was found. If closed, it will open the connection , get the item, and close it again. If you plan to use myset[] "massively", then call myset.load()

Tip: myset[] can also accept myset['_tsId=TS01']. Useful when the unique value is not only the id.

fonsecareyna82 (approved):

PR #370, opened 2022-10-14
Fix import volumes

pconesa:
Code separated EMDB vs File
logging used.
Possible outputs defined.
Url to get EMDB metadata updated. The previous one was failing.
More consistent when failing. Now protocol also fails.


rmarabini (approved):
OK with me.  The tests seem to work ;-)

PR #372, opened 2022-10-28
tolerate absence of tifffile (case for devel intallations) triggering…

pconesa:
… all the import cascade when running pip install -e path/to/scipion-em from the setup.py.

### DIFF
--- a/pwem/__init__.py
+++ b/pwem/__init__.py
@@ -32,6 +32,7 @@ from pkg_resources import parse_version
 
 import pyworkflow as pw
 from pyworkflow.protocol import Protocol
+from pyworkflow.utils import weakImport
 from pyworkflow.viewer import Viewer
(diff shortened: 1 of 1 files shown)


azazellochg:
Hi! I'm not sure I understand why tifffile is absent in devel case..

azazellochg:
Ah, is it because it is in the __init__?

fonsecareyna82 (approved):

pconesa:
When installing in production pip first reads the dependencies and install them. Then all is fine when is em installation turn.

In devel mode we are doing pip install -e pat/to/devel

Which enters the setup.py, and then, import __init__ ....and find tifffile as first missing dependency in the "casacade import"

If the the installer we do:

pip install -r requirements.txt first this wouldn't happen.

PR #379, opened 2022-11-24
* Fixing an error related with the import coordinates(Closes #375)

### LINKED ISSUE #375: Bug import coordinates
https://github.com/scipion-em/scipion-em/blob/1d0beb11ea4606be50776c81fdb730a50481fb71/pwem/protocols/protocol_import/coordinates.py#L228

Hi, 

This line will make "mic_100_coords.start" match with mic_1.mrc, which is wrong.

I guess that the proper way of dealing with it will be to compare the maximum common substrings.

This is my dirty patch

```
    def getMatchingMic(self, coordFile, fileId):
        """ Given a coordinates file check if there is a micrograph
        that this files matches.
        """
        from difflib import SequenceMatcher
        micSet = self.inputMicrographs.get()

        if fileId is None:
            coordBase = pwutils.removeBaseExt(coordFile)
            matcher = SequenceMatcher(a=coordBase)
            best_match_size = -1
            best_match_micId = None
            for mic in micSet:
                micBase = pwutils.removeBaseExt(mic.getFileName())
                matcher.set_seq2(micBase)
                match = matcher.find_longest_match(alo=0, ahi=len(coordBase), blo=0, bhi=len(micBase))
                if match and match.size > best_match_size:
                    best_match_size = match.size
                    best_match_micId = mic.getObjId()
            if best_match_micId:
                #TODO: Update a cache to avoid searching for the best_match micrograph in the next getMatchingMic call
                return micSet[best_match_micId]
            else:
                return None
        else:
            return micSet[fileId]

```

fonsecareyna82:


pconesa (changes requested):
Thanks Yunior! You are missing updating CHANGES.txt ;-)

What about avoiding using getUniqueFileName in the imports.... will it come later?

pconesa:
@fonsecareyna82 we need to addredd the Scipion iport case too here:
https://github.com/scipion-em/scipion-em/blob/devel/pwem/protocols/protocol_import/dataimport.py#L138

pwem/__init__.py
```diff
@@ -89,6 +89,9 @@ class Config(pw.Config):
     # Path to either ImageJ or Fiji binary program
     IMAGEJ_BINARY_PATH = _get("IMAGEJ_BINARY_PATH",'',description="Path to the IMAGEJ or FIJI program.", var_type=VarTypes.PATH, source="pwem")
 
+    SCIPION_EM_NEW_FILE_CHECK_SEC = int(_get('SCIPION_EM_NEW_FILE_CHECK_SEC', 10))
+    "Number of seconds to wait before checking if new files are available or finished in streamified import protocols."
```
pconesa, line 93: Could you please, add the description as an argument and the source and type.

pwem/protocols/protocol_import/images.py
```diff
@@ -94,7 +94,7 @@ def _acquisitionWizardCondition(self):
 
     # --------------------------- INSERT functions ----------------------------
     def _insertAllSteps(self):
-
+        self._fillFileNamesList()
```
pconesa, line 97: This mfails in come tests.... I think those that overwrite _insertAllSteps and therefore does not call _fiilFileNameList

 scipion3 tests pwem.tests.protocols.test_protocols_sets.TestSets

pconesa:
one more thing, hopefully, the last one. Could yo please address the sonarcloud issue (duplicated code)

pconesa (approved):

PR #382, opened 2023-01-09
fix particle import

rmarabini:
Let us assume the following situation:

Two data sets (datasetA and datasetB) were acquired and processed independently. Then, we want to merge the particles and process the whole dataset together.

By default the files containing the set of particles are mic001.mrcs, mic002.mrc, etc. That is, there is file mic0001.mrcs for datasetA and another with the same name for datasetB. Since both are in different directories joining both sets of particles is not a problem. At this point we decide that the project is too large and we import the merged setOfPArticles in a new empty project. The protocol "import particles" creates a link for each binary file, the local (in the import file protocol) name is the basename of the file. For example the following two links are created

ln -s Runs/datasetA/extra/mic0001.mrcs Runs/importParticles/extra/mic0001.mrcs
ln -s Runs/datasetB/extra/mic0001.mrcs Runs/importParticles/extra/mic0001.mrcs

That is, the link mic0001.mrcs is created twice in the protocol_import directory. Each time the link point to a different file and therefore, the result is that no link is created to datasetA because is overwritten by the link created to datasetB

The proposed solution is to stop using the basename as local name and change it by an slug, that is:
instead of

ln -s Runs/datasetA/extra/mic0001.mrcs Runs/importParticles/extra/mic0001.mrcs

the protocol will create

ln -s Runs/datasetA/extra/mic0001.mrcs Runs/importParticles/extra/runs-datasetA-extra-mic0001.mrcs

  this is all


azazellochg:
Removing traceability from Scipion is going well..

pconesa:
We(@fonsecareyna82 and me) have been looking at this case, not only for particles but for any import.

The best we think we can do is:

Try to keep the basename (mic0001) unless we can. In that case, we will, prepend the folders (losing all traceability) But in this case there is no way we can keep traceability having duplicated names.

rmarabini:
> We(@fonsecareyna82 and me) have been looking at this case, not only for particles but for any import.
> 
> The best we think we can do is:
> 
> Try to keep the basename (mic0001) unless we can. In that case, we will, prepend the folders (losing all traceability) But in this case there is no way we can keep traceability having duplicated names.

Looks like a reasonable solution for my and I am happy with it. 
Nevertheless, I do not understand why prepending the path decreases the traceability. I see that the name of the links will not be identical to the names of original images. Instead, it will be the name of the original image plus a prefix, clearly no information is lost. Furthermore, the function that creates the new filenames would fix some problematic filenames since it removes spaces and converts non ASCII characters to the closer ASCII ones.
Again, I am happy with the solution proposed by @pconesa and @fonsecareyna82

pconesa:
Could you please test if your case is covered here --> https://github.com/scipion-em/scipion-em/pull/379

pconesa:
I think using 2 images with the same name looses the traceabiltiy in itself, rules we want to do something with the folders containing the images, which makes it more complex.

rmarabini:
Hola Pablo,

    No se muy bien cómo probar esto. Instale scipio usando

pip3 install --user scipion-installer

de forma que estoy usando el pwem que hay en

~roberto/miniconda/envs/scipion3/lib/python3.8/site-packages/pwem/

si hago un clone del repositorio scipion-em, y selecciono  la rama
fix_import_coordinates
¿cómo le digo a Scipion que use el código nuevo? por mas que intento instalarlo
scipion usa el código que hay en
~roberto/miniconda/envs/scipion3/lib/python3.8/site-packages/pwem/

   gracias

    Roberto

rmarabini:
Hi @pconesa 

   This pull request does not solve the problem I described when importing particles. 
In the following lines you may see the number of files after importing with this pull request (14321) and after importing properly (19442)

ls -lrt 004631_ProtImportParticles/extra/ | wc
  14321  157522 2778092
ls -lrt 004192_ProtImportParticles/extra/ | wc
  19442  213853 4646411

pconesa:
I had time to see why the other PR request does not fix this and al is here:
https://github.com/scipion-em/scipion-em/blob/devel/pwem/protocols/protocol_import/dataimport.py#L138

This code, should be adapted in the same way is done for the rest of the import protocols. Will be addressed in the other PR. I'll leave it in draft until is solved and hopefully close this one

pconesa:
Let's resolve this, A recent code already in devel is supose to address this case more thoroughly. 

After all this time the solution I propose is:

1.- close this PR (@rmarabini thank very much to this contribution anyway)
2.- Rescue the "sequence.py" part (Is it still valid?) Current code in devel remains the same evaluating None).
3.- JAVA_MAX_MEMORY: I think 10 GB as the default value seems a lot to me. ¿what do you think?

rmarabini:
Hi,

  I guess sequence.py is valid (at least I have not modify it in a long while)
If you have several millions of particles then JAVA_MAX_MEMORY=10Gb is needed. May be with the new viewer this requirement is not longer needed.

pconesa:
Ok, I'll prepare a new branch, just with the sequence.py stuff. The new viewer does not need this value and should deal with 1M particles fine.
________________________________
De: rmarabini ***@***.***>
Enviado: miércoles, 19 de junio de 2024 11:48
Para: scipion-em/scipion-em ***@***.***>
Cc: Pablo Conesa ***@***.***>; Mention ***@***.***>
Asunto: Re: [scipion-em/scipion-em] fix particle import (PR #382)


Hi,

I guess sequence.py is valid (at least I have not modify it in a long while)
If you have several millions of particles then JAVA_MAX_MEMORY=10Gb is needed. May be with the new viewer this requirement is not longer needed.

PR #384, opened 2023-01-09
averageMovie: removed as it's no longer used

DStrelak:
See also https://github.com/I2PC/scipion-em-xmipp/pull/603

### DIFF
--- a/pwem/protocols/protocol_align_movies.py
+++ b/pwem/protocols/protocol_align_movies.py
@@ -573,56 +573,6 @@ class ProtAlignMovies(ProtProcessMovies):
         runJob(self._log, eman2.Plugin.getProgram(program), args,
                env=eman2.Plugin.getEnviron())
 
-    def averageMovie(self, movie, inputFn, outputMicFn, binFactor=1, roi=None,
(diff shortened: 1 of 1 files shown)


pconesa:
Just now I realized that this method is used.... here:

https://github.com/scipion-em/scipion-em-grigoriefflab/blob/ae26dad18fc21503ced373f63d64e6ddb3838750/grigoriefflab/protocols/protocol_unblur.py#L232

@azazellochg , xmipp team is reviewing and cleaning xmipp code. The are removing "xmipp_movie_alignment_correlation", therefore we are loosing averageMovie method. Used in "unblur" to do de PSD......

Is removing the option to of the PSD and option?

azazellochg:
@pconesa grigoriefflab plugin has been deprecated years ago

DStrelak:
> Just now I realized that this method is used.... here:
> 

That usage can be easily replaces by `xmipp_image_statistics --save_image_stats`

pconesa:
upsss! unblurr is not in cistem anymore?

azazellochg:
unblur is in cistem. it does not use averageMovie

pconesa (approved):

PR #389, opened 2023-01-24
UniProt typo and alphabet initialization for export

DaniDelHoyo:
When the sequence object is initialized, if no alphabet is defined it is set as None, but this breaks the construction of the SequenceHandler. We should either initialize the Alphabet to some value when creating the Sequence object if not provided or at least do not use the None for SequenceHandler.

### DIFF
--- a/pwem/convert/sequence.py
+++ b/pwem/convert/sequence.py
@@ -162,13 +162,13 @@ class SequenceHandler:
         error = ""
         if dataBase is None:
             if self.isAminoacid:
-                dataBase = 'UnitProt'
+                dataBase = 'UniProt'
             else:
                 dataBase = 'GeneBank'
 
         while counter <= retries:  # retry up to 5 times if server busy
             try:
-                if dataBase == 'UnitProt':
+                if dataBase == 'UniProt':
                     url = "http://www.uniprot.org/uniprot/%s.xml"
                     format = "uniprot-xml"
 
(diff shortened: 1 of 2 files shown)


rmarabini:
Hi @DaniDelHoyo,

I would prefer your second option that is, check if iUPACAlphabet is None in SequenceHandler constructor

could you check if adding "or iUPACAlphabet is None"  solves your problem?

```
class SequenceHandler:
    def __init__(self, sequence=None,
                 iUPACAlphabet=Alphabet.DUMMY_ALPHABET):
        ###if iUPACAlphabet == Alphabet.DUMMY_ALPHABET:  ### OLD LINE
        if iUPACAlphabet == Alphabet.DUMMY_ALPHABET or iUPACAlphabet is None:  ### NEW LINE
            self.isAminoacid = None
        else:      
            self.isAminoacid =  (iUPACAlphabet <   Alphabet.AMBIGOUS_DNA_ALPHABET)
```

    Thanks

DaniDelHoyo:
It does solve the problem yet. This is what you asked for right?

rmarabini (approved):

rmarabini:
Hi @DaniDelHoyo and @pconesa 

   This pull request is OK with me I do not merge it in devel branch  because I think Pablo should have  the last word.

DaniDelHoyo:
Can we merge this then @pconesa ?

DaniDelHoyo:
actually not, I just realized it fails somewhere else too

PR #392, opened 2023-01-31
Rm sequence

rmarabini:
Added testLog method to capture error messages from Phenix

If it is a fatal error, for example when map and model do not overlap an exception is raised and the protocol fails. (Before the protocol did not fail although no valid output was produced. ) 
If the error is not fatal the protocol continues its execution printing a warning. Usually this second case is related with format inconsistencies in the input file that the protocol try to fix using maxit.
---
We have also added to the function readFromPDBDatabase a counter. Now the function will retry the download "counter" number of times if an error occurs during the connection.


pconesa (approved):

pconesa:
@rmarabini , could you please add a single lie in CHANGES.txt summarizing this changes?

rmarabini:
@pconesa 

done
V3.0.26
   - function readFromPDBDatabase tries to download an atomic structure up to 5 times before returning  an error (before it tried only once)
   - function retry search for fatal error messages in the output and if found,  it aborts instead of trying several times.

PR #400, opened 2023-03-09
New protocol boxsize checkpoint

danielmarchan3:
Protocol to compare two particles box size


pconesa (changes requested):
Good Job Dani! protocols, conf, tests .... ;-)

You only are missing a line in changes.txt. PLease add it and all will be ok.

pconesa (approved):

PR #404, opened 2023-03-23
 Keeping original base names

fonsecareyna82:
* Keeping original base names
        - changing only those that are repeated
  * Removing python-slugify dependency


rmarabini (approved):
Seems to work ;-)

PR #409, opened 2023-04-06
Jmrt pwem

delarosatrevin:
## Two main changes in this PR:
* Refactor some movie align functions to be used in a batch implementation of motioncor2
* Created ProtSubsetRange for selecting a subset (e.g. 1-50k) of element from a set. 
   *  Although this functionality is implemented in the more general ProtSubset protocol, it is extremely SLOW
   * For example, selecting 50k subset out of 300k input set, took 3h in ProtSubet, the same operation is only 8s in ProtSubsetRange


azazellochg (approved):
movie alignment changes look good to me. I leave subsets for @pconesa

pwem/protocols/protocol_sets.py
```diff
@@ -884,3 +884,84 @@ def _summary(self):
...
+        end = (rangeTo if rangeTo >= 0 else inputSet.getSize()) - 1
+
+        for i, item in enumerate(inputSet.iterItems()):
+            if i > end:
+                break
+            if i >= start:
+                outputSet.append(item)
+
+        if n := outputSet.getSize():
+            key = 'output' + inputClassName.replace('SetOf', '')
```
pconesa, line 942: With this type of protocols, (subsets, joins,...) It is almost impossible to define possible outputs with this naming scheme. My suggestion would be to choose a fixed name and define possible outputs so it can be used when planning a workflow.

pwem/protocols/protocol_sets.py
```diff
@@ -884,3 +884,84 @@ def _summary(self):
...
+
+    # -------------------------- DEFINE param functions -----------------------
+    def _defineParams(self, form):
+        form.addSection(label='Input')
+
+        add = form.addParam  # short notation
+        add('inputSet', pwprot.params.PointerParam, pointerClass='EMSet',
+            label="Input set of items", important=True,
+            help='Input from where the subset will be created.')
+        add('rangeFrom', pwprot.params.IntParam, default=1,
```
pconesa, line 906: I like more the idea of the String param and specifying the id like 1-200, 250-300 ...

It is more flexible...not sure if this is precisely the origin of the performance issue in the "traditional subsets" protocol?

pconesa (commented):
NIce improvement from 3h to 8s!! Would be nice to understand why this difference.

I've commented some lines. Cheers.

pwem/protocols/protocol_sets.py
delarosatrevin: Hi Pablo, as you said, it is almost impossible...since it can create subsets of many set types. So, I prefer to not define any possible output because it will add many "possible" entries to the selection dialog. I think this protocol is mostly used through API, where the workflow can be built anyway...or more in normal GUI processing one step after the other.

pconesa:
Haven't got time to review. I'll try to find some time this week.

pwem/protocols/protocol_sets.py
```diff
@@ -571,81 +571,48 @@ def createOutputStep(self):
...
+        if nElements > 100000:  # show progressBar for large sets
+            progress = ProgressBar(total=nElements, fmt=ProgressBar.NOBAR)
+            progress.start()
+            sys.stdout.flush()
+            step = max(25000, nElements // 25000)
 
-            for finalId in finalIds:
+        i = 0
 
-                item = inputFullSet[finalId]
```
pconesa, line 647: Good job Jose Miguel... I think this way of accessing the set  set[id] is a performance killer.

pconesa (approved):

PR #421, opened 2023-05-22
Fix setVolume in AS object

DaniDelHoyo:
AS volumes cannot be set with Volume subclasses like Subtomogram Averages.
Would it make sense to allow it?


rmarabini (approved):
OK with me

pconesa (approved):

PR #424, opened 2023-06-07
Convert symmetries

pconesa:


rmarabini (changes requested):
@pconesa please add a test for the new method >:-)

By the way I have modify the description of the new class/mwethods

pconesa:
The initial test (#50) is using already the new code.

rmarabini (approved):

PR #433, opened 2023-07-04
Flexibility Hub attributes inside Scipion objects

DavidHerreros:
Flexibility Hub attributes have been added to standard Scipion objects.

Method `copyInfo` and subsets have also been modified to copy extended attributes,

### DIFF
--- a/pwem/objects/data.py
+++ b/pwem/objects/data.py
@@ -573,6 +573,8 @@ class Image(EMObject):
         self._origin = None
         if location:
             self.setLocation(location)
+        self._zFlex = None
+        self._zRed = None
 
     def getSamplingRate(self):
         """ Return image sampling rate. (A/pix) """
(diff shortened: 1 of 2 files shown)


pwem/objects/data.py
```diff
@@ -759,6 +764,23 @@ def setMRCSamplingRate(self):
         """ Sets the sampling rate to the mrc file represented by this image"""
         from pwem.convert.headers import setMRCSamplingRate
         setMRCSamplingRate(self.getFileName(), self.getSamplingRate())
+    def getZFlex(self):
```
pconesa, line 767: Are all these needed by particle and volume?

pwem/objects/data.py
```diff
@@ -1295,6 +1318,14 @@ def hasCTF(self):
     def setHasCTF(self, value):
         self._hasCtf.set(value)
 
+    def hasFlexInfo(self):
+        return bool(self._flexProg.get())
```
pconesa, line 1322: This will fail when self._flexProg is None

pconesa:
please, confirm this is not needed anymore @DavidHerreros

DavidHerreros:
Hi @pconesa 

Yes, these changes would not be needed anymore as now the data model of FlexHub is inside Scipion. We can close this PR.

PR #435, opened 2023-07-28
Raise exception when input is empty

DaniDelHoyo:
Just to handle wizard errors when empty input

### DIFF
--- a/pwem/wizards/wizards.py
+++ b/pwem/wizards/wizards.py
@@ -472,6 +472,8 @@ class SelectAttributeWizard(VariableWizard):
             item = inputSet.getFirstItem()
         elif issubclass(inputSet.__class__, pwobj.Object):
             item = inputSet
+        else:
+            raise ValueError('The input is not defined')
         return item
 
     def getInputAttributes(self, form, inputParam):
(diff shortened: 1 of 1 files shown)


pwem/wizards/wizards.py
```diff
@@ -472,6 +472,8 @@ def getFirstItem(self, form, inputParam):
             item = inputSet.getFirstItem()
         elif issubclass(inputSet.__class__, pwobj.Object):
             item = inputSet
+        else:
+            raise ValueError('The input is not defined')
```
pconesa, line 476: Shouldn't we capture the exception at getInputAttributes and show a message instead of a generic exception message?

DaniDelHoyo: I don't know about doing it in the show, you have to access to the Pointer first, check if it is a PointerList and in that case taking the first and finally checking if it is empty. Quite some operations that would need to be repeated.
We might also reconfigure the wizard so the inputSet is got in the show with a "getInputSet" function and checking there if it is empty, but then we would need to modify a lot of child wizards.
About the error type, I am open to change it yes, how should we make the message?

pconesa: So, the call stack would be like:

Main thread ...
   wizard.show ...
       getInputAttributes ....
             getFirstItem  ....


The new raise method will reach "Main thread" and will show some kind of error.

I think it is better to capture the exception in the show and then, use showInfo or showError method to show the proper message

pconesa (approved):

PR #445, opened 2023-08-28
Bug fix: Dimension errors with new copyInfo

DavidHerreros:
When copyInfo is called with a `SetOfParticles`, the new copy call will also add the image dimension information. In protocols changing these dimensions (i.e. doing a cropping or resizing), the previous behaviour results in a wrong registration of the new image dimensions.

Therefore, the attribute `_firstDim` has been added to the ´ignoreAttrs´ list to fix the previous issue.

### DIFF
--- a/pwem/objects/data.py
+++ b/pwem/objects/data.py
@@ -1377,7 +1377,7 @@ class SetOfImages(EMSet):
         """ Copy basic information (sampling rate and ctf)
         from other set of images to current one"""
         if type(self) is type(other):
-            self.copy(other, copyId=False)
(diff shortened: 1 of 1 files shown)


pconesa:
This should have been fixed already in a different way. https://github.com/scipion-em/scipion-em/pull/440/files

Were you updated?

DavidHerreros:
The machine was not updated... sorry for that! I will try again after updating Scipion

DavidHerreros:
After updating everything works fine, thank you so much Pablo! I close the PR.

PR #449, opened 2023-10-03
classifyItems method: updates an existing output sets

danielmarchan3:
- This can be used for streaming protocols such as xmipp PCA streaming classification


CHANGES.txt
```diff
@@ -1,3 +1,7 @@
+V3.3.1:
+    Developers:
```
pconesa, line 2: There is no need to increse de version 3.3.0 will be the next. Move this to the 3.3.0 section please

danielmarchan3: Okay

pconesa (approved):

PR #452, opened 2023-10-11
Pc symetry plot refactorization

pconesa:
This is, for now what I've got for the angular distribution viewer.

offers 3 plots: 2d, 3d and heatmap and to chose the colormap

image


pconesa:
Need to optimize ploting (search for weights) and restrict plot to n elements?

rmarabini:
I am looking to this now...

azazellochg:
A few comments on the plots:

- plot 1 has no axes labels. What is weight and how it is decided and applied? Could you swap min and max colors (max = dark blue, min = white)? Does 35 mean 35 particles?
- plot 2 is not very good, I can't tell anytthing from it. Could we have a button to open chimerax bild file instead?
- plot 3 rot angle is now +/-30 range, while on plot 1 you use 0-360 - inconsistent convention. Could we also swap white and blue? What is 0.5? Weight? Why it does not match plot 1?

azazellochg:
Also, what do the color scale options high/low mean in the viewer?

azazellochg:
Thanks Pablo! 
For plot1 axes, we could label tilt angle,e.g. https://stackoverflow.com/questions/35498450/matplotlib-polar-plot-set-label-position
I think you missed my question about weight. For a person who never used scipion/xmipp, what does it suppose to mean?

rmarabini (changes requested):
Hi @pconesa ,

    Follows a few comments, as you can see this angular projection distribution raises passions.
 
1) It should be clear what is displayed in each plot So I suggest to add to the title the following labels.
a) "Euler angles" for the cryosparc like plot
b) "projection direction angles" for the polar plot
c) projection directions for the 3D plot

For the plot (b) I would suggest to make clear what is used as radial coordinate and what is used as angular coordinate.  My guess is that polar angle is used for the radius and the azimuthal angle for the rotational angle (see definiton at  https://en.wikipedia.org/wiki/Spherical_coordinate_system)

Personally the plot I find more informative is (c) -in this I believe I disagree with Grigory- since it has a direct geometric interpretation.

What I miss in plot (C) is 
   (1) where is the origin? that is the angles (0,0) are plot at....((0,0,1) or (1,0,0) or (0,1,0). A red line from the center of the sphere to the point with angles (0,0) would solve this problem
   (2) my tests, with a few millions of particles, are painfully slow so I suggest these modifications 
to improve speed [I am focusing now about plot (c) but can be generalized]
      2.1] create a uniform distribution of points in the sphere surface and calculate the weight in this surface. The point here is that the present calculation of weight is not efficient and has a  complexity of O(n**2). A possible way to create such sampling is the fibonacci_sphere algorithm (see attached file fibonacci_sphere.py) 

[fibonacci_sphere.py.txt](https://github.com/scipion-em/scipion-em/files/13039596/fibonacci_sphere.py.txt)

    2.2] Since cos and sin functions are used a lot I suggest to create a lookup table (see file memorization.py). In my tests a table with 10,000 entries provides the proper results up to 4 decimal digits.

[memoerization.py.txt](https://github.com/scipion-em/scipion-em/files/13039619/memoerization.py.txt)

  2.3] In order to calculate the weight a serial search is performed. Instead of a serial search perform binary search in the variable "z" (as defined by fibonnaci) and once z has been found we need to look for the right "theta" (as defined by fibonnanci)  moving plus/minus 180/ga angles (ga=golden angle~2degree)

An example of binary search may be found in file  binary.txt

[binary.py.txt](https://github.com/scipion-em/scipion-em/files/13039669/binary.py.txt)

I am sure this email is a bit confusing but I am more than willing to meet @pconesa  and discuss the details

  Roberto

pconesa:
@rmarabini , can we find some time to address those issues together?

pconesa:
This is how it currently looks for a icosahedral refinement

image

azazellochg:
My 50 cents:
plot 1) can we have either gray bg or larger point size, you can't see white dots on white. 
plot 2) I guess the purpose of this fake 3D plot which is actually 2D is to show the symmetry? Otherwise it's pretty much the same as plot 1
plot 3) text is messed up; probably symmetrical around 0 makes more sense. And again this weight, which should be replaced by # of ptcls.

azazellochg:
I'd say we should have both cryosparc 2D and relion/chimera-bild 3D plots at a minimum, if you want to attract any users of these software to Scipion. I don't mind adding extra plots, but the main two should be present.

pconesa:
Thanks @azazellochg , 
Good point about white dot over white background.... just one note.... the viewer offers the chance to chose the color map, it doesn't have to be the one I chose for the screenshot....but it is true that is the default one (not a good default one)


I think you are not getting 3d plot. It is not faked it is an actual 3d plot of the orientations... Maybe the fact that they are restricted to a symmetry unit cell is misleading you?.... a c1 symmetry will cover half of the sphere.

Regarding plot 3 (CS).... I agree that units are not right but we did not find a good solution when looking at the code...let me have another try...maybe just setting the color scale range would work.

Regarding the scale.... I think there is no problem to shit the values.... but actual euler angles (what the plot is ploting) ranges from -180 to 180 that is why we left them untouched. And again a c1 refinement will cover the whole area the same way is done with CS.

azazellochg:
Hi @pconesa for the fake 3d I meant it is projected to 2d, or you can actually rotate this plot in 3d? For plot 3 scaling - I guess it's fine if these are raw values. For the units, I have a script that does this:

image

pconesa:
Thanks @azazellochg I'll have a look and the script.... and YES, it is a matplotlib 3d plot. It is like chimera one with rotation and interactivity

azazellochg:
That's awesome!! Then my comments are irrelevant :)


Best regards,
Grigory

--------------------------------------------------------------------------------
Grigory Sharov, Ph.D.

MRC Laboratory of Molecular Biology,
Francis Crick Avenue,
Cambridge Biomedical Campus,
Cambridge CB2 0QH, UK.
tel. +44 (0) 1223 267228 <+44%201223%20267228>
e-mail: ***@***.***

rmarabini:
Hi @pconesa,

   I have added the test file  "pwem/tests/protocols/test_angular_distribution.py"  to the pull request. Since the target is a viewer,  the tests are no automatic and require a person to visualize and check them.  

As far as I am concerned this pull request is ready to go.

pconesa:
i'm merging this. All seems good. The only part I think we could improve is the CS plot, to make it as mentioned by Grigory.

> https://stackoverflow.com/questions/35498450/matplotlib-polar-plot-set-label-position

PR #461, opened 2023-11-21
     - Fix: CTFTomoSeries manual subset should work

pconesa:
- Windows' ChimeraX properly called from WSL installations
     - EMSet.copyItems and createCopy accepts "itemSelectedCallback" falling back to isEnabled.
     - getVersionFromPath improved: works with cuda-11.4.safe (uses regex looking up for numbers)

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,12 +1,15 @@
 V3.4.0:
    Users:
      - Improving performance of the metadataviewer.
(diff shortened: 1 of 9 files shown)


azazellochg:
getVersionFromPath is working now, thank you @pconesa !

JorMaister (approved):

PR #462, opened 2023-11-21
Rm wait

rmarabini:
added protocol "wait". This protocol sleeps for N seconds (value provided by de user) and then ends. The wait protocol is useful to schedule the start of other protocols using the "wait for" feature.

Why I find the procotol wait useful? Many protocols allow the use of a scratch disk. That  is, the data is copied before the protocol execution from a slow (magnetic) disk to a faster (solid state) disk. If two process (data copies) access at the same time to the magnetic disk the performance drops dramatically. So I :
   (1) launch  the data hungry protocol A with the  scratch disk option on and  get a estimation of the copy time
   (2) start the wait protocol with that number of seconds
   (3) launch the second data hungry protocol B with "wait for" pointing to wait

This approach is much, much more faster than start both protocols at the same time and let then compete for the hard disk.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,3 +1,7 @@
+V3.5.0
+  Users:
+     - add wait protocol. This protocol wait N seconds and then ends.
+       It is helpfull to schedule other protocols using the option "wait for"
 V3.4.0:
    Users:
      - Improving performance of the metadataviewer.
--- a/pwem/__init__.py
+++ b/pwem/__init__.py
@@ -42,7 +42,7 @@ from .objects import EMObject
 from .tests import defineDatasets
 from .utils import *
 
-__version__ = '3.4.0'
+__version__ = '3.5.0'
 NO_VERSION_FOUND_STR = "0.0"
 CUDA_LIB_VAR = 'CUDA_LIB'
 
--- a/pwem/protocols/__init__.py
+++ b/pwem/protocols/__init__.py
@@ -67,4 +67,4 @@ from .protocol_boxsize_checkpoint import ProtBoxSizeCheckpoint
 from .protocol_mathematical_operator import ProtMathematicalOperator
 
 from .protocol_tools import *
-
+from .protocol_wait import ProtWait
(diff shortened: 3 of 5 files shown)


CHANGES.txt
```diff
@@ -1,3 +1,7 @@
+V3.5.0
```
pconesa, line 1: We are still in time tio include this for 3.4.0 if you want?

pconesa:
One question.... why not waiting for "protocolA". Is it because the hungry scratch process is only partial to the whole protocol A?

rmarabini:
Hi @pconesa 

> One question.... why not waiting for "protocolA". Is it because the hungry scratch process is only partial to the whole protocol A?

Let us assume that I want to execute relion-3D-classify (this will be protocol A)

The protocol is divided in 3 steps:
   1) convert
   2) execute a long program
   3) create output

If a scratch disk is used, the data is copied at the very beginning  of step (2) just before the classification starts.
I want protocol B to be started as soon as the copy of the data to the scratch disk have finished in protocol A.
I do not want to wait till the classification is done. Therefore wait untill Protocol has finished is not an option.

That is, I do not want both protocols to access at the same time to the magnetic disk but it is OK to access to the scratch disk.

hope this helps to clarify the situation

pconesa (approved):

delarosatrevin:
Instead of creating a new type (dummy?) protocol, I would have added an expert option to the general protocol class, that allows any protocol to wait N seconds before starting.

rmarabini:
I do not disagree with this suggestion. In fact it was my first option but I thought that other "scipioners"  will complain if I made a modification in the base. @pconesa  what do you think?

pconesa:
I think Jose Miguel's idea and yours is better

delarosatrevin:
@rmarabini I think it could be very unintrusive, an extra option that will never be considered if not passed. Regarding the GUI, we could "reuse" the Wait box, which is intended for waiting for another protocol, we could, for example, detect if the value there is ending in an s (seconds) and then activate that wait option.

pconesa:
I'd avoid refusing the wait field for different meanings. It will
complicate things there it is used.

 Actually the wait field should have a magnifier to look for protocols.

rmarabini:
@pconesa. I see your point with the wait field.

PR #481, opened 2024-02-22
Coords viewer

fonsecareyna82:
* Adding a new 2D coordinate viewer.
* Updating the classifyItems method:
    - adding a boolean flag indicating whether to raise an exception if there is a failure while attempting to retrieve the next element 
       from itemDataIterator.

*  Updating the STKImage Reader:
    - reading a .stk file without the id@ ...

### DIFF
--- a/pwem/viewers/__init__.py
+++ b/pwem/viewers/__init__.py
@@ -41,10 +41,12 @@ from .viewer_sequence import SequenceViewer
 from .viewer_volumes import viewerProtImportVolumes
 from .viewer_angular_dist import AngularDistributionViewer
 from .viewers_data import BasicMDViewer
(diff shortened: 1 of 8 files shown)


pwem/viewers/__init__.py
```diff
@@ -41,10 +41,12 @@
 from .viewer_volumes import viewerProtImportVolumes
 from .viewer_angular_dist import AngularDistributionViewer
 from .viewers_data import BasicMDViewer
+from .viewer_coordinates import CoordinateViewer
 
 try:
     from .mdviewer import MDViewer
 except Exception as e:
+    print(e)
```
pconesa, line 49: I don't think we want this, right?

PR #499, opened 2024-04-29
Extract coordinates: modify ending condition

danielmarchan3:
This protocol would not finish if both micrographs and ctf sets where closed, however there is a possibility that you were using trigger points so you dont want to use all your data, when this happened even though the input coordinates set was closed it had to wait until all the micrographs and ctfs that where not associated to those coordinates to finish. Now it will check if the input coords are closed and if you have all the ctfs and mics from the input already processed to finish without waiting for all to finish.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,3 +1,10 @@
+v3.8.0
+  Users:
+    - Allows to run the whole workflow until 3D processing with trigger points
+
+   Developers
+    - Modify the finishing condition of the parent protocol "extract particles". Now, it checks if the input coords are closed and if all mics and CTFs have been processed.
+
 V3.7.0:
   Users:
      - Allowing drag and drop a coordinate when picking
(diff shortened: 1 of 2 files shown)


CHANGES.txt
```diff
@@ -1,3 +1,10 @@
+v3.8.0
```
pconesa, line 1: Next version is 3.7.0. Please add you nice detailed line to the 3.7.0 section? ;-) Unless we want to wait for the next release?

pwem/protocols/protocol_particles.py
```diff
@@ -313,7 +313,16 @@ def _useCTF(self):
 
     # ------ Methods for Streaming extraction --------------
     def _isStreamClosed(self):
-        return self.micsClosed and self.ctfsClosed and self.coordsClosed
+        return self.coordsClosed
+
+    def _isAllMicsProcessed(self):
```
pconesa, line 318: _areAllMicsProcessed ?

danielmarchan3: True it was a misspelling, I already change it.

pconesa (approved):

PR #517, opened 2024-07-14
reduce number of getDim calls

azazellochg:


pconesa (commented):
Wow! All this time this has been here! Nice catch!

What about "lazy getting" dims. Something like:

def  getDim(self):

    if self._dim is None:
        #self._dim = #store the dimensions
        
    return self._dims


Will this work?

There might be a risk of not getting the right dimensions when changing the filename/location?

azazellochg:
It might be too risky as you say. Such change will need much more investigation

azazellochg:
Shall we merge it?

delarosatrevin:
Good fix @azazellochg ! 

Not sure if related to this, but at the createOutputStep of a Relion 2d, I'm getting a lot of calls retrieving the dimensions.

```
Getting the dimensions for the first item: /lustre_scratch/user_scratch/jdela80/scipion/20240719_Krios02_braf_craf_kd_s3_OTF/Runs/000387_ProtRelionExtractParticles/extra/mic_000017.mrcs
00909:   Getting the dimensions for the first item: /lustre_scratch/user_scratch/jdela80/scipion/20240719_Krios02_braf_craf_kd_s3_OTF/Runs/000387_ProtRelionExtractParticles/extra/mic_000017.mrcs
00910:   Getting the dimensions for the first item: /lustre_scratch/user_scratch/jdela80/scipion/20240719_Krios02_braf_craf_kd_s3_OTF/Runs/000387_ProtRelionExtractParticles/extra/mic_000017.mrcs
00911:   Getting the dimensions for the first item: /lustre_scratch/user_scratch/jdela80/scipion/20240719_Krios02_braf_craf_kd_s3_OTF/Runs/000387_ProtRelionExtractParticles/extra/mic_000017.mrcs
```

Repeated many times for one stack and also for several stacks.

azazellochg:
thanks! Every protocol need to be investigated :)

pconesa:
That comment is being there for months. And actually the idea is to reveal "abuse" of getDim calls.   As I saw, the use of getX, getY and getZ implies a call each time... Maybe this is the case? I'd recommend to use x,y,z = obj.getDims() instead.

pconesa (approved):

azazellochg:
Please keep this comment, it's very useful indeed!

PR #522, opened 2024-08-05
    - Chimera viewer compatible with newer versions via linking

pconesa:


azazellochg (approved):

pconesa:
let me check, thanks!

pconesa:
seems Version package does not like 11.2.safe anymore?

azazellochg:
I think what you need to do is change the test to use "guessCudaVersion" instead, which will print a new warning and set a default version.

pconesa:
Yes I can improve the test

pconesa:
Should work now.... still it could be better to have a regular expression  but should cover most cases.

PR #524, opened 2024-08-07
Move "from Bio import pairwise2" to the methods that use it so the de…

JorMaister:
…precation warning when opening Scipion is hidden


pconesa (approved):

pconesa:
With these changes, we are hiding the depreciation warning at loading time, but this alignment should be addressed.

Would you @mmmtnez, @rmarabini or @DaniDelHoyo have time to look at it.

Will this help? https://www.biostars.org/p/9553992/

DaniDelHoyo:
As far as I have checked, my code does not contain that import so I have no problem hiding it.
There are indeed some programs I use that actually rely on that function, but I think I cannot help it, I also hope they update their codes at some point.

PR #539, opened 2024-10-07
Enabled editing nested sets in "edit set" protocol

oierlauzi:
Added support to edit nested sets (e.g. SetOfClasses, SetOfTiltSeries...) in the "edit set" protocol.


oierlauzi:
Done

> Please, add a line in CHANGES.txt

pconesa:
Got a better solution for this. Thanks anyway. Formula for subitems.

PR #543, opened 2024-10-23
skip creating class2D is classId=0

azazellochg:
Relion intermediate iterations can have particles with classId=0 that do not have any class assignment yet. These should be skipped when creating classes

### DIFF
--- a/pwem/objects/data.py
+++ b/pwem/objects/data.py
@@ -2296,6 +2296,8 @@ class SetOfClasses(EMSet):
                 ref = newItem.getClassId()
                 if ref is None:
                     raise Exception('Particle classId is None!!!')
+                if ref == 0:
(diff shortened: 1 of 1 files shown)


pwem/objects/data.py
```diff
@@ -2296,6 +2296,8 @@ def classifyItems(self,
                 ref = newItem.getClassId()
                 if ref is None:
                     raise Exception('Particle classId is None!!!')
+                if ref == 0:
```
pconesa, line 2299: I guess this means that None is an invalid class and raises an exception but 0 means that we need to skip the class and the particle, right?

azazellochg: correct!

pconesa (approved):

PR #549, opened 2024-11-22
biopython 1.81 can not longer access PDB atomic structures, this is a…

rmarabini:
… quick fix that can be reverted as soon as biopython 1.84 is the default version


azazellochg:
@rmarabini it would be great if you could fix the issues raised by sonarcloud

rmarabini:
Hi @azazellochg, 

    Let me comment the issue I try to solve with this pull request. In Scipion, the protocol that imports atomic models from PDB had stop working. It seems that the PDB ftp server is not longer maintained and the only way to download  files is using http. Scipion uses biopython (version 1.81) to handle the downloads and this version  uses ftp. The first version of biopython that addresses this problem is biopython 1.84 that requires python==3.9 while scipion uses python==3.8. That is, we can not update biopython.

What I did was to copy a biopython file named pdb_list.py from biopython module to scipion and modify it so now it works with the http server (the modification is a single line), then I forced Scipion to use this modified file instead of the one in the biopython module.  In summary, sonar cloud is complaining about biopython code. I will delete all code that is not used right now by scipion (this will fix the security issues) but certainly I do not plan to modify biopython code style or conventions

pconesa (approved):

PR #569, opened 2025-06-09
Do not assume chimera plugin has binaries

martin-s-a:
In the GitHub Actions of scipion-em-kiharalab, we are seeing an error lately where the scipion-em protocol ProtImportPDB fails, and the reason we found is that, as our requirements file includes scipion-em-phenix, which itself includes scipion-em-ccp4, which then includes scipion-em-chimera, this last plugin is being installed with our plugin, but not the binaries for chimera.

This means that, when reaching [this line](https://github.com/scipion-em/scipion-em/blob/af16d763fb2434363f51631b507d4ec80670fcbd/pwem/protocols/protocol_import/volumes.py#L373), the import will work, as the plugin exists and can be imported, but then the operation using chimera fails because of the lack of the binaries.

To solve this, I have created a function `__getChimeraPlugin` that returns the plugin when exists and the binaries are present, or `None` otherwise, so the operation using chimera only takes place if both conditions are satisfied.

### DIFF
--- a/pwem/protocols/protocol_import/volumes.py
+++ b/pwem/protocols/protocol_import/volumes.py
@@ -373,7 +373,6 @@ Format may be PDB or MMCIF"""
 
         chimeraPlugin = self.__getChimeraPlugin()
         if chimeraPlugin:
-            from chimera import Plugin as chimeraPlugin
             localPath = localPath[:-4] + localPath[-4:].replace(".pdb", ".cif")
             args = f'--nogui --cmd "open {atomStructPath}; save {localPath}; exit"'
             chimeraPlugin.runChimeraProgram(chimeraPlugin.getProgram(), args)


pwem/protocols/protocol_import/volumes.py
```diff
@@ -415,6 +417,15 @@ def _validate(self):
                           self.pdbFile.get())
         # TODO: maybe also validate that if exists is a valid PDB file
         return errors
+    
+    @staticmethod
+    def __getChimeraPlugin():
```
pconesa, line 422: Thanks Martin, makes perfect sense.

What about improving the naming?

__getChimeraPlugin actually does more than that:

What about namig this method:

isChimeraAvailable ? Returning true or false?

martin-s-a: The thing with `isChimeraAvailable` is that, if true, then you have to then run the import again, something the function checking its existance has already done. So, the goal of `getChimeraPlugin` is to already retrieve a ready-to-use module instance (or None in case it does not exist) to avoid duplicating code.

martin-s-a: It depends on if you prefer to check the existance explicitly in the function at the cost of duplicating the import, vs seeing it as an object retrieval function instead of condition evaluation, without the code duplication.

In my personal opinion, the latter is better, but it's that, an opinion.

pconesa: then remove line 376

martin-s-a: Oops, right, line 376 is redundant

martin-s-a: Done

pconesa (approved):

PR #571, opened 2025-06-12
Added protocol to filter items by tilt angle 

oierlauzi:
Very frequently users need to filter SetOfXXX to obtain top-views or side-views. I usually instruct them to use filter set with a very long formula. This protocol aims to streamline this operation in a user friendly way,


pwem/objects/data.py
```diff
@@ -152,7 +152,7 @@ def getRotationMatrix(self):
     def getEulerAngles(self):
         from pwem.convert.transformations import euler_from_matrix
         rotation = self.getRotationMatrix()
-        return euler_from_matrix(rotation)
+        return euler_from_matrix(rotation, 'szyz')
```
pconesa, line 155: May be better to be an argument "axes" with 'szyz' by default sois more versatile?

pwem/protocols/protocol_filter_by_normal.py
```diff
@@ -0,0 +1,141 @@
...
+        errors = []
+
+        inputSet = self._getInputSet()
+        item = inputSet.getFirstItem()
+        try:
+            self._getMatrix(item)
+            
+        except (AttributeError, TypeError, ValueError):
+            errors.append('Input items must have a valid transform associated. '
+                          'Try modifying "Matrix Field" parameter if necessary')
```
pconesa, line 118: I guess, that not all items have _transform, thus the field "Matrix Field".  Wouldn't be better to scan attributes to find the transform one... otherwise, how the user is goingto find the attribute name?

PR #574, opened 2025-07-17
Save non import errors

DaniDelHoyo:
At some point, I found that chimera raises some errors when trying to open a atomstruct and save it as cif. 
Therefore, this structure would fail.
I have modify it so regardless of the error, if Chimera cannot be used, the file is just copied

### DIFF
--- a/pwem/protocols/protocol_import/volumes.py
+++ b/pwem/protocols/protocol_import/volumes.py
@@ -372,10 +372,11 @@ Format may be PDB or MMCIF"""
 
         try:
             from chimera import Plugin as chimeraPlugin
-            localPath = localPath[:-4] + localPath[-4:].replace(".pdb", ".cif")
-            args = f'--nogui --cmd "open {atomStructPath}; save {localPath}; exit"'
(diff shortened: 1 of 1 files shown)


DaniDelHoyo:
This new change makes the ChimeraView look for existing chimera folders in the EMROOT with a bit more entusiasm

rmarabini (changes requested):
I think findHome should be something like (that is, no chimera-1.2.5 hard encoded): 

import os

    @classmethod
    def findHome(cls):
        """Return 'chimerax' if it exists, otherwise the newest chimerax-XXXXXX version."""
        files = [f for f in os.listdir(emConfig.EM_ROOT) if os.path.isdir(os.path.join(emConfig.EM_ROOT, f))]
        files.sort(reverse=True)

        # Case 1: exact 'chimerax' directory exists
        if "chimerax" in files:
            return os.path.join(emConfig.EM_ROOT, "chimerax")

        # Case 2: fallback to latest chimerax-XXXXXX
        for f in files:
            if f.startswith("chimerax"):
                return os.path.join(emConfig.EM_ROOT, f)

        # Case 3: nothing found
        return None

PR #581, opened 2025-09-05
Fixing Euler angle convention in getEulerAngles()

oierlauzi:
In CryoEM we usually use the `szyz` convention for Euler Angles. Transform.getEulerAngles() returns them in the `sxyz` convention. Currently the function is not used much across plugins: https://github.com/search?q=org%3Ascipion-em%20getEulerAngles&type=code

### DIFF
--- a/pwem/objects/data.py
+++ b/pwem/objects/data.py
@@ -149,10 +149,10 @@ class Transform(EMObject):
         M = self.getMatrix()
         return M[:3, :3]
 
-    def getEulerAngles(self):
+    def getEulerAngles(self, axes='szyz'):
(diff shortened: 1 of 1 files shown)


pconesa (changes requested):
Mayb better offer "axes" in getEulerAngles, using by default th szyz convention?

JorMaister:
Agreed, @pconesa

oierlauzi:
Probably the last element of the euler angles (which is the only one used in tomo) is probably numerically the same between both conventions. Let me check it.

pconesa (approved):

oierlauzi:
@pconesa please, revert this PR as it is breaking tomo

PR #588, opened 2025-11-10
Creation of new wizard to select more than one chain in an atom structure (SelectChainSWizard)

mmmtnez:
The new wizard SelectChainSWizard is a copy of the previous existing one SelectChainWizard, which allowed to select only one chain of an atom structure, just modified to select more than one chain.

Additionally, appropriate indentation has been included in the next SelectResidueWizard wizard.


rmarabini (approved):
seem to be OK

